PR #4, opened 2020-05-31
Request: add php7-fileinfo

noahwsmith:
This is a relatively common dependency for filesystem manipulation- "The functions in this module try to guess the content type and encoding of a file by looking for certain magic byte sequences at specific positions within the file. While this is not a bullet proof approach the heuristics used do a very good job."


nigelgbanks:
Looks good to me

PR #6, opened 2020-06-03
Openseadragon working with cantaloupe 4

dannylamb:
## What does this PR do?

- Updates cantaloupe.properties to work with Cantaloupe ^4.0
- Configures tomcat to allow url encoded slashes for Cantaloupe
- Exposes Cantaloupe at drupal.localhost/cantaloupe instead of cantaloupe.localhost to dance around CORS
- Enables the traefik dashboard for debugging
- Pulls down latest openseadragon in Dockerfile
    - Deploys it in the startup script since its location is on a bind mount

## How to test?
- Make a Repository Item and give it a display hint of "Openseadragon"
- Give it an Image media tagged as Original File
- Go view the node and you should have an OSD viewer with zoom/scroll

## Caveats

I can't seem to upload tiffs appropriately.  This doesn't have anything to do with cantaloupe as far as I can tell.  I can upload it in the widget in Drupal, but after I save the form, a File entity is created but the file itself is not in Fedora.  :man_shrugging:  Looks like it's getting stripped out by Traefik or Fedora's nginx proxy.  I'll defer to @nigelgbanks on that one.

### DIFF
--- a/cantaloupe/rootfs/etc/confd/templates/cantaloupe.properties.tmpl
+++ b/cantaloupe/rootfs/etc/confd/templates/cantaloupe.properties.tmpl
@@ -2,6 +2,7 @@
 # GENERAL SETTINGS
 ###########################################################################
 
+temp_pathname = {{ getv "/temp/pathname" "" }}
 http.enabled = {{ getv "/http/enabled" "true" }}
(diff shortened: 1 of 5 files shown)


nigelgbanks:
Don't know what the problem is with tiff files, but that's definitely a different pull request / ticket.

Issue #21, opened 2020-06-04
Building on Windows 10 fails


marksandford:
./gradlew.bat build failing on Windows 10 Pro, build 19041 with message:

```
> Task :imagemagick:build FAILED

FAILURE: Build failed with an exception.`

* What went wrong:
Execution failed for task ':imagemagick:build'.
> Process 'command 'docker'' finished with non-zero exit value 1
```

$ java -version
java version "1.8.0_251"

$ docker-compose version
docker-compose version 1.25.5, build 8a1c60f6
docker-py version: 4.1.0
CPython version: 3.7.4
OpenSSL version: OpenSSL 1.1.1c  28 May 2019

$ docker -v
Docker version 19.03.8, build afacb8b

At Danny's suggestion, I enabled experimental features, but that didn't seem to help.

[Console output with --stacktrace](https://github.com/Islandora-Devops/isle-buildkit/files/4732055/gradlew_build.txt)

nigelgbanks:
@marksandford could you also add run the following command and attach the output to this ticket.

```bash
docker info
```

marksandford:
@nigelgbanks 
here you go:

$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 25
 Server Version: 19.03.8
 Storage Driver: overlay2
  Backing Filesystem: <unknown>
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc version: dc9208a3303feef5b3839f4323d9beb36df0a9dd
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.19.76-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 1.944GiB
 Name: docker-desktop
 ID: H5KL:MFWP:EUQK:HCVT:JZIN:LVQW:4SH4:U6O5:J6RW:RWB6:OM37:3G6T
[...]

PR #24, opened 2020-06-06
Build issues on Windows 

nigelgbanks:
Should address the issue: https://github.com/Islandora-Devops/isle-buildkit/issues/21

Address two issues:
- Previously Windows EOL characters were used in some files not covered by .gitattributes.
- Previously download files that did not download successfully would pollute the build cache.

Solutions:
- Change .gitattributes so the appropriate files have the appropriate line endings.
- For all download files check against a known checksum, if check fails remove the file from the build cache so it can be downloaded again.


nigelgbanks:
@marksandford Mind giving this a test to confirm if it solves your issues?

PR #25, opened 2020-06-08
Matomo

dannylamb:
This installs Matomo if no pre-existing database exists.  The company that made matomo specifically doesn't want you to do non-interactive installs (they offer it as a service i think).  So this is the "accepted practice"... dumping an existing db and passing it around :vomiting_face: So yeah... I template out a sql file and import it.  Just avert your :eyes:  FWIW it's what we're doing in the playbook, too.

Other than that, there were a few gotchas.  I'm adding a X-Forwarded-Uri header with traefik so things work out ok with Matomo behind the proxy.  I also changed the "front porch" url to `islandora.localhost`, and gave Traefik and alias of `islandora.localhost` so the backend services can still reference each other using their public urls.  We could've just kept going with `drupal.localhost`, but there is a difference between the drupal service and "islandora itself", so I thought it'd be better to be explicit.

## Testing Instructions

- Pull in this PR
- ./gradlew build
- docker-compose up -d
- visit `islandora.localhost/matomo/` and you should be routed to matomo's login
- login with admin/password
- you got a matomo site!
- go visit `islandora.localhost` and do some stuff
- go back to `islandora.localhost/matomo/` and now you should see your visit in the dashboard


docker-compose.yml
```diff
@@ -137,14 +133,11 @@ services:
       - fcrepo
       - database
       - activemq
-    networks:
-      default:
-        aliases:
-          - drupal.localhost
```
nigelgbanks, line 143: Does Gemini/Recast and derivative generation still work without an alias?

docker-compose.yml
```diff
@@ -234,5 +233,7 @@ services:
       # Do not expose in production.
       - traefik.http.routers.api.service=api@internal
     networks:
-      - external
-      - default
+      external:
+      default:
+        aliases:
+          - islandora.localhost
```
nigelgbanks, line 239: Oh I see never mind my comment above.

nigelgbanks:
On a quick review this looks good to me, I'll test in the morning and if all is well, I'll merge it.

nigelgbanks:
Works locally, I needed to destroy my existing volumes but that's fine.

PR #32, opened 2020-06-11
Redirect all logs to stdout/stderr, and some formatting.

nigelgbanks:
Useful for services/systems that output to single aggregate like AWS


nigelgbanks:
Should be discussed at the meeting today, if this is the route we want to go.

nigelgbanks:
Was discussed at the meeting yesterday and approved.

dannylamb:
Huge changeset but I can follow it.  Let me give it a whirl and I'll follow up.

Issue #30, opened 2020-06-10
Matomo bind mount / network mount support 


nigelgbanks:
In trying to get [EFS](https://aws.amazon.com/efs/) as a persistent volume type under ECS requires that we do an initial copy into the NFS mount as it doesn't behave like like normal volumes. This applies as well for bind mounts or any volume drivers than the default.

nigelgbanks:
Resolved by https://github.com/Islandora-Devops/isle-buildkit/pull/45

PR #33, opened 2020-06-11
Matomo no longer needs a volume as the configuration is generated by …

nigelgbanks:
…the startup scripts.

Should address Issue: https://github.com/Islandora-Devops/isle-buildkit/issues/30


nigelgbanks:
Will setup to support bind mount

PR #44, opened 2020-06-12
Remove default volumes, now specify exclusively in docker-compose, etc

nigelgbanks:
All the volumes that have been removed are already defined in the docker-compose.yml file.


dannylamb:
Rebuilding :rocket:

PR #46, opened 2020-06-12
Removed volume for Karaf as it does not appear to be needed.

nigelgbanks:
As far as I can tell we do not need to keep changes to the /opt/karaf/data directory at runtime between restarts, we just need what is put in there by installing bundles, etc.


docker-compose.yml
```diff
@@ -83,10 +82,6 @@ services:
       - traefik.http.routers.activemq_http.entrypoints=http
   alpaca:
     image:  ${REPOSITORY:-local}/alpaca:latest
-    volumes:
-      - karaf-data:/opt/karaf/data
-    depends_on:
-      - activemq
```
dannylamb, line 89: Don't we want to keep this?

nigelgbanks: It doesn't really have much of an affect aside from when running ``docker-compose up alpaca`` causes `activemq` to start as well, it doesn't guarantee startup order or anything. I don't think we need it.

PR #51, opened 2020-06-15
Use plugin for build logic.

nigelgbanks:
Moved the Gradle logic to it's own plugin so it can be shared across projects.


dannylamb:
Man that's nice.  Testing it out now. :rocket:

settings.gradle.kts
```diff
@@ -1,4 +1,11 @@
-rootProject.name = "docker"
+rootProject.name = "isle-buildkit"
+
+// Include plugin for building Docker images.
+sourceControl {
+    gitRepository(uri("https://github.com/Islandora-Devops/isle-gradle-docker-plugin.git")) {
```
dannylamb, line 5: Is the assumption here that if I needed to bugfix the plugin, I could check it out locally and change this uri to my filesystem somehow?  Just trying to figure out the workflow if there's an issue.

nigelgbanks: If you wanted to use a path on disc to the project instead for the purposes of bugfixing you could use this following instead of the `sourceControl` bit.

```kotlin
includeBuild("../isle-gradle-docker-plugin")
```

Assuming the folder where it is checked out is up one directory.

Issue #9, opened 2020-06-04
Fedora - Add support for multiple backends to fedora


g7morris:
_Originally from the https://github.com/Islandora-Devops/isle-buildkit/README.md_

* Add support for multiple backends to fedora (currently only file is being used)

* This functionality to toggle between backends (file, mysql and postgres) already exists in [isle-dc](https://github.com/Islandora-Devops/isle-dc/blob/development/fcrepo.env#L34) & within the [isle-fcrepo image](https://github.com/Islandora-Devops/isle-fcrepo/blob/dev/rootfs/opt/fcrepo/config/fcrepo-config.xml#L42) as a feature which should be ported to `isle-buildkit`.

* Is Fedora configured to use [Blazegraph](https://github.com/Islandora-Devops/isle-buildkit/issues/7)?

nigelgbanks:
I'm taking a look at this now as I'm trying to get s3 working.

dannylamb:
We now support mysql and postgresql.  With Fedora 6 those are the only two options (other than the embedded h2 for testing purposes), so we're good to close this one.

PR #65, opened 2020-07-07
Add support for postgres where possible, and now support database bot…

nigelgbanks:
…h database for fcrepo.

This is significantly bigger pull as I've refactored as much of the database logic down into the base image so it can be shared by all the services.

This required changing the defaults from 'database' to the specific implementation since 'matomo' does not work with postgresql we can't assume only one database in the stack (if users choose to use postgres) so the defaults have changed to 'mariadb'.

To be tested along with https://github.com/Islandora-Devops/isle-dc/pull/86


nigelgbanks:
Address https://github.com/Islandora-Devops/isle-buildkit/issues/9

dannylamb:
@nigelgbanks Can you resolve the conflicts on this?  Normally I'd do it myself but Github's not letting me through the UI.

nigelgbanks:
@dannylamb will do

nigelgbanks:
@dannylamb ready to go

dannylamb:
Testing now :rocket:

PR #73, opened 2020-09-04
Fix moved key

nigelgbanks:
Looks like main was created from non master branch, so the commits that were in master have not been pulled in? @dannylamb 
Let me know if I should rebase this off main and if the 'main rename' commits need to wait before going into main.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -2,7 +2,7 @@ name: CI
 on: 
   push:
     branches:
-      - master
+      - main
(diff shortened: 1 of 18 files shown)


nigelgbanks:
This is basically cause the previous key server is a bit unreliable, so I've switched it to the one that was most recommended.

dannylamb (approved):

dannylamb:
It's ok, we'll pull all those changes in right here.

PR #75, opened 2020-09-14
Fixes display issues with some JP2 images.

nigelgbanks:
Had an issue where in cantaloupe tried to use Kakadu even though it wasn't installed so switched selection to manual. Also made sure that OpenJPEG tools were installed so they could be used appropriately. I think previous calls must have been using the ImageMagick JPEG2000 delegate which is a bit slower. See docs: https://cantaloupe-project.github.io/manual/4.1/images.html

Can test the existing implementation vs the changes against the attached JP2. Prior to this pull request it does not render, afterwards it does.

[Sample.zip](https://github.com/Islandora-Devops/isle-buildkit/files/5218114/Sample.zip)

### DIFF
--- a/cantaloupe/Dockerfile
+++ b/cantaloupe/Dockerfile
@@ -8,7 +8,7 @@ RUN --mount=type=cache,target=/var/cache/apk \
     apk-install.sh \
         imagemagick \
         ffmpeg \
-        openjpeg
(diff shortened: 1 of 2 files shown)


dannylamb (approved):
Cool.  We've had JP2 issues for a while now.  Hope this helps.

PR #76, opened 2020-09-30
Allow parameterization of wait_for_service function

emetsger:
Support waiting for specified services, which is useful when optional services like `fcrepo` and `blazegraph` aren't used.  Otherwise, `wait_for_service` will timeout.


nigelgbanks (approved):
Looks safe to me

PR #77, opened 2020-10-01
Fix demo startup.

nigelgbanks:
Potentially temporary fix for change introduced https://github.com/Islandora/controlled_access_terms/pull/54.

Which removed the dependency on the `name` module by splitting the features configuration into install/optional and allowing users to manually select in the Features UI to not install optional dependencies. Unfortunately `drush features:import` doesn't expose this functionality. So for now we do not import those features which is the same route the playbook took: https://github.com/Islandora-Devops/islandora-playbook/pull/190/files.


seth-shaw-unlv:
We needed to fix the playbook (https://github.com/Islandora-Devops/islandora-playbook/pull/190) but I didn't even think about ISLE needing a similar fix.

dannylamb:
...and now I'm magically running into this.  Testing now.

dannylamb:
@seth-shaw-unlv A lot of what we do mirrors the playbook here.  But a lot of that is gonna go when we change things so you build your own Drupal container, though. So for the time being they're kinda linked.

dannylamb:
We're good.  This gets the init script completing again.

dannylamb (approved):

PR #78, opened 2020-10-01
Matomo support multisites

nigelgbanks:
Adds multi-site support for Matomo.

Changes:

- Check more keyserver pools for when using gpg to make builds a bit more stable when keyservers do not respond.
- Change Matomo nginx setup to support hosting a with/without `/matomo` path prefix.
- Create database entries per configured site, and update if variables change.
- Make sure all multi-sites are considered trusted hosts.


nigelgbanks:
Was a couple of commits behind main so I rebased.

nigelgbanks:
Was a couple of commits behind main so I rebased.

dannylamb:
Getting this when going to visit Matomo:
image

The database dump that gets templated out needs to respect the `$DOMAIN` variable we're setting now.

matomo/rootfs/etc/confd/templates/setup-environment.sh.tmpl
```diff
@@ -17,11 +17,30 @@ foreground {
   MATOMO_DB_ROOT_PASSWORD="{{ getv "/db/root/password" "password" }}"
   MATOMO_DB_ROOT_USER="{{ getv "/db/root/user" "root" }}"
   MATOMO_DB_USER="{{ getv "/db/user" "matomo" }}"
-  MATOMO_SITE_HOST="{{ getv "/site/host" "islandora.isle-dc.localhost" }}"
-  MATOMO_SITE_NAME="{{ getv "/site/name" "Islandora" }}"
-  MATOMO_SITE_TIMEZONE="{{ getv "/site/timezone" "America/Halifax" }}"
+  MATOMO_DEFAULT_SITE_HOST="{{ getv "/default/site/host" "islandora.isle-dc.localhost" }}"
```
dannylamb, line 20: This default should get changed to `islandora.traefik.me`, though i suppose that'll get covered in your other PR to set up the defaults.

dannylamb:
Oh... maybe it's just an issue with the `docker-compose.matomo.yml` file in isle-dc.  MATOMO_SITE_HOST should become MATOMO_DEFAULT_SITE_HOST.  Let me try that, and if it's good, i'll merge this.

matomo/rootfs/etc/confd/templates/setup-environment.sh.tmpl
nigelgbanks: I'll do that in the other pull for sure, as I'll have to rebase it anyways after this is merged and make some changes.

nigelgbanks: Hmm though might be easier to just set the default here then the isle-dc pull request can wait.

dannylamb:
Yep, I made those changes and now we're good:
image

PR #79, opened 2020-10-07
Add GNU patch for use with composer-patches.

emetsger:
Adds GNU `patch` to the Drupal container in order to support the use of `composer-patches`, and attempts to follow the existing pattern/usage of buildkit directives.  I can revise the PR if you prefer this to occur in a different image, or if you want to incorporate the commands into a single layer, just LMK.

### Background
https://github.com/cweagans/composer-patches is a module which allows for local patches to be applied to vendored code required by `composer.json`.  Johns Hopkins uses `composer-patches` to apply local patches to SimpleSAMLphp.  The version of `patch` present on the alpine image isn't compatible with `composer-patches`, thus the need for this PR.

### DIFF
--- a/drupal/Dockerfile
+++ b/drupal/Dockerfile
@@ -1,6 +1,11 @@
 # syntax=docker/dockerfile:experimental
 FROM local/nginx:latest
 
+RUN --mount=type=cache,target=/var/cache/apk \
+    apk-install.sh \
+      patch && \
+    cleanup.sh
+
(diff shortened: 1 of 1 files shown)


nigelgbanks (approved):

nigelgbanks:
Seems like a good addition, most likely only composer in the Drupal projects will make use of it so that image should be as good as any. If someone later has a use case for using patch in another image another pull could be raised to move it into the base image.

PR #80, opened 2020-10-13
Using secrets

dannylamb:
This approach lets you use secrets for running in production.  For an environment variable that should be secret, instead of directly entering its value into config, replace it with `secret:/path/to/the/secret`.  This little init script will scan the container environment for variables that follow that pattern, read the contents of that file, and replace the variable's value with the file's contents.

There will be an analagous pull on isle-dc to demonstrate how to use secrets in the docker-compose file.  I'm trying to work this in such that if integrating with other things like etcd, vault or aws's secrets solution, that this won't get in the way. It's specifically for the env backend of confd and if you don't pattern your variables this way, nothing will change.

### DIFF
--- a/base/rootfs/etc/cont-init.d/00-load-secrets.sh
+++ b/base/rootfs/etc/cont-init.d/00-load-secrets.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/with-contenv bash
+set -e
+
+# Read any secret files specified in environment variables.
+echo "$(env | grep '=secret:')" | while read line
+do
+    # Skip empty lines
+    [[ -z $line ]] && continue
+
+    # Hack out the path to the secret.
+    environment_variable=$(echo $line | cut -d= -f1)
+    secret=$(echo $line | cut -d= -f2 | cut -d: -f2)
+
+    # Load the secret's value into the environment variable
+    if [ -f ${secret} ]; then
+        s6-env -i ${environment_variable}="$(cat ${secret})" s6-dumpenv -- /var/run/s6/container_environment
+    fi
+done
--- a/demo/Dockerfile
+++ b/demo/Dockerfile
@@ -22,7 +22,6 @@ RUN --mount=type=cache,target=/root/.composer/cache \
         drupal/facets:^1.3 \
         drupal/matomo:^1.7 \
         drupal/pdf:1.x-dev \
-        drupal/rdfui:^1.0-beta1 \
         drupal/rest_oai_pmh:^1.0 \
         drupal/restui:^1.16 \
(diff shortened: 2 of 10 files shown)


demo/Dockerfile
```diff
@@ -22,7 +22,6 @@ RUN --mount=type=cache,target=/root/.composer/cache \
         drupal/facets:^1.3 \
         drupal/matomo:^1.7 \
         drupal/pdf:1.x-dev \
-        drupal/rdfui:^1.0-beta1 \
```
nigelgbanks, line 25: I assume this is fine, just thought I'd bring attention to it if it was intended for another branch or commit.

nigelgbanks (approved):

base/rootfs/etc/cont-init.d/00-load-secrets.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/with-contenv bash
+set -e
+
+# Read any secret files specified in environment variables.
+echo "$(env | grep '=secret:')" | while read line
```
noahwsmith, line 5: Could you put instructions into the README? It looks like you have to do a specific thing with a filename to make this work, and we shouldn't assume people will go read the code to figure it out.

noahwsmith: And can you add your SECRETfile as a template to this repo?

noahwsmith: If you scroll to "By .env technique" here you can see how to load a second .yml file by changing your COMPOSE_FILE declaration in your .env file https://stackoverflow.com/questions/55650342/import-docker-compose-file-in-another-compose-file

dannylamb:
I've made an associated pull to isle-dc: https://github.com/Islandora-Devops/isle-dc/pull/123

I'll try to document it in the README as best I can here.

dannylamb:
I decided it fit better in isle-dc than here.  I updated the PR to include a secrets chunk in the README.

base/rootfs/etc/cont-init.d/00-load-secrets.sh
dannylamb: I put example secret files in https://github.com/Islandora-Devops/isle-dc/pull/123

dannylamb:
@nigelgbanks Any chance we can get this in to help with https://github.com/Islandora-Devops/isle-dc/pull/126?

g7morris:
@dannylamb per your request here https://github.com/Islandora-Devops/isle-dc/pull/126#issuecomment-726818269 I merged this PR

PR #81, opened 2020-10-15
Update allowed-external-content.txt.tmpl

dannylamb:
During development I noticed the fcrepo service was logging warnings about whitespace in the `allowed-external-content.txt` file. Turned out to be a newline at the end of the file. This just shuffles the end of the loop in the template, which prevents a newline from getting added.  Apparantly it's a golang template thing.


nigelgbanks:
Nice find!

PR #86, opened 2020-11-05
Allow Drupal to lock tables, allows importing dumps generated by Drus…

nigelgbanks:
…h: https://www.drupal.org/project/drupal/issues/2487089.


dannylamb:
:+1: Surprised I haven't bumped into this.  I guess I've always just dumped using mysqldump directly when playing around with it.

dannylamb (approved):

PR #87, opened 2020-11-19
Injecting DRUPAL_REVERSE_PROXY_IP for traefik if it's available

dannylamb:
I've managed to get the https situation sorted out a bit better, and this lets us properly configure drupal for a reverse proxy setup.  In order to make things work before it needed to be misconfgured on purpose, so this is a bit of an improvement :sweat_smile: 

If you're not using Traefik, you have to provide the IP(s) of the reverse proxy to Drupal via `DRUPAL_REVERSE_PROXY_IPS` using a comma separated list of IPs.  If you are using Traefik, it will try and auto-detect it generate the variable for you to save a step.

In order to actually set the settings, there's a new drush command, `islandora:settings:set-reverse-proxy`

You'll need this for an upcoming `isle-dc` PR and can test it with that.

### DIFF
--- a/drupal/rootfs/etc/confd/templates/setup-environment.sh.tmpl
+++ b/drupal/rootfs/etc/confd/templates/setup-environment.sh.tmpl
@@ -2,8 +2,22 @@
 # -*- mode: sh -*-
 # vi: set ft=sh:
 with-contenv
+# If traefik is not found allow any IP address.
+backtick -D 0.0.0.0 -n TRAEFIK_IP {
+  redirfd -w 2 /dev/null
+  backtick -i -n CAPTURE {
+    getent hosts traefik
+  }
+  importas CAPTURE CAPTURE 
+  pipeline {
+    echo ${CAPTURE}
+  }
+  awk "{ print $1 }"
+}
 multisubstitute
 {
+  # Non-site specific variables
+  importas REVERSE_PROXY_IPS TRAEFIK_IP
   # Default settings to apply if none given.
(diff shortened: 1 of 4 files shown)


nigelgbanks:
Not sure that this needs to be configurable via an environment variable, the requesting address is available as a global variable in `php`, for example in the following snippet, could be used instead:

```php
// Required when running Drupal behind a reverse proxy.
$settings['reverse_proxy'] = TRUE;
$settings['reverse_proxy_addresses'] = array($_SERVER['REMOTE_ADDR']);
$settings['reverse_proxy_trusted_headers'] = \Symfony\Component\HttpFoundation\Request::HEADER_X_FORWARDED_ALL;
```
As the address will be passed on via `nginx` from the request origin which may be `traefik` or an `alb` from AWS. Since the Drupal container itself would not be directly accessible from any other source other than the proxy I don't think using `$_SERVER['REMOTE_ADDR']` represents a security threat.

nigelgbanks (changes requested):

nigelgbanks:
I've raised a pull request against your branch here: https://github.com/Islandora-Devops/isle-buildkit/pull/88 with what I think might work, testing will tell.

nigelgbanks:
The pull request I raised is not sufficient. It allows the variable to be set but a reasonable working default is still not possible.

dannylamb:
:bowing_man: I merged the execline, many thanks. Is there still something left for me to do here @nigelgbanks ?

nigelgbanks (approved):

nigelgbanks:
Good to go I think

PR #88, opened 2020-11-20
Allow DRUPAL_REVERSE_PROXY_IPS to be set explicitly via whatever back…

nigelgbanks:
…end is configured for confd.

### DIFF
--- a/drupal/rootfs/etc/confd/templates/setup-environment.sh.tmpl
+++ b/drupal/rootfs/etc/confd/templates/setup-environment.sh.tmpl
@@ -2,8 +2,22 @@
 # -*- mode: sh -*-
 # vi: set ft=sh:
 with-contenv
+# If traefik is not found allow any IP address.
+backtick -D 0.0.0.0 -n TRAEFIK_IP {
+  redirfd -w 2 /dev/null
+  backtick -i -n CAPTURE {
+    getent hosts traefik
+  }
+  importas CAPTURE CAPTURE 
+  pipeline {
+    echo ${CAPTURE}
+  }
+  awk "{ print $1 }"
+}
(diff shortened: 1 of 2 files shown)


nigelgbanks:
Hey @dannylamb here are some of the changes I suggested on your pull request, let me know what you think.

nigelgbanks:
I have only tested the script and not actually tested the affect of using 0.0.0.0 for the reverse proxy as a default if `traefik` is not available. I'll post a comment after I've confirmed it to be working as expected.

nigelgbanks:
`0.0.0.0` does not work... and since we can't insert code with `drupal_rewrite_settings` where a bit out of luck.

dannylamb:
Yeah, it's tricky.  What I did was just how I could figure out how to hack that in.  

The execline stuff scares me a bit :ghost:  At least now I can kinda see what you're doing there.

I have no idea what the best solution to all this is.  The default is bound to be different if you're running a docker-compose file from `isle-dc` vs something like AWS.  And the $_SERVER var is different b/c we're in drush and not a page load.  I'm trying to autodetect, but very ham-fistedly.  :man_shrugging:

nigelgbanks:
I've raised https://github.com/Islandora-Devops/isle-buildkit/issues/89 to get ride of the `execline` code as it can go die in a fire. So consider it temporary.

nigelgbanks:
`$_SERVER['REMOTE_ADDR']` is fine with Drush I'm actively been using it for quite some time.

dannylamb:
If we can pass that through as a default i'll be happy.

dannylamb (approved):

dannylamb:
Swinging back around to this

PR #91, opened 2020-12-04
Update Nginx Proxy cache to avoid 502 errors on heavy content pages

emudojo:
When loading certain pages that are packed with  heavy content/LONG headers nginx crashes with a 502 error, this specially true when trying to edit nodes

https://vander.host/knowledgebase/web-servers/502-bad-gateway-upstream-sent-too-big-header-while-reading-response-header-from-upstream-on-hosted-nginx-website/

this patch fixes that and also forces SSL on ALL resources

### DIFF
--- a/base/rootfs/etc/cont-init.d/00-load-secrets.sh
+++ b/base/rootfs/etc/cont-init.d/00-load-secrets.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/with-contenv bash
+set -e
+
+# Read any secret files specified in environment variables.
+echo "$(env | grep '=secret:')" | while read line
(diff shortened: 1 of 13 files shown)


nigelgbanks:
@emudojo There doesn't seem to be any file changes on the pull request is there something else to be added?

noahwsmith:
It was trying to do this: https://github.com/Born-Digital-US/isle-buildkit/commit/5a575ec0e6d8fd3e73f165442b05eac10aed450c#diff-ffa457e827e71e4569ace093e80ded1de69af789a8692cdab7a91d55c1c53939R48

PR #95, opened 2020-12-08
Sandbox, Import/Export tool, and Syn toggling

dannylamb:
Does the above^^

Plus it guts the startup script and I'm pushing that stuff into the Makefile in `isle-dc`.  Will link PR when it's ready.


drupal/rootfs/etc/confd/templates/setup-environment.sh.tmpl
```diff
@@ -33,7 +33,7 @@ multisubstitute
   define DB_USER "drupal_default"
   define EMAIL "webmaster@localhost.com"
   define INSTALL "true"
-  define INSTALL_EXISTING_CONFIG "false"
+  define INSTALL_EXISTING_CONFIG "true"
```
nigelgbanks, line 36: Maybe override it in the demo image as there is no configuration in the base Drupal image.

demo/Dockerfile
```diff
@@ -5,56 +5,22 @@ FROM local/nginx:latest as composer
...
-    composer create-project drupal/recommended-project:^8.9 \
-        --prefer-dist \
-        --no-interaction \
-        --stability stable \
-        --no-dev \
-        --no-install \
-        /var/www/drupal \
-    && \
+    # Get the Drupal codebase
+    git clone https://github.com/dannylamb/islandora-sandbox && \
```
nigelgbanks, line 9: There is a helper https://github.com/Islandora-Devops/isle-buildkit/blob/main/crayfish/Dockerfile#L8 `git-clone-cached.sh` which might be helpful in reducing build times.

noahwsmith:
The only thing I found is that GEMINI needs to be remove here:

https://github.com/Islandora-Devops/isle-buildkit/blob/post-sprint-4/drupal/rootfs/etc/islandora/utilities.sh#L80

dannylamb:
:bowing_man: Thx for the commit.

I think it's time this becomes a real pull request.

dannylamb:
Missed a default.  I've confirmed this works with and without secrets now.

noahwsmith:
Pulling all the images into our work, this seems to work well without secrets. I must have missed something in my merge because secrets don't work anymore with all of this merged into our work. But they seem to work independently...

dannylamb:
@noahwsmith I think I changed one of the secrets from being Fedora specific to Tomcat specific.  Maybe it's that?  I'm pretty sure I can revert that and we can test.

fcrepo6/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+| Environment Variable           | Etcd Key                        | Default                           | Description |
+| :----------------------------- | :------------------------------ | :-------------------------------- | :---------- |
+| FCREPO_ACTIVEMQ_BROKER         | /fcrepo/activemq/broker         | tcp://activemq:61616              |             |
+| FCREPO_ACTIVEMQ_QUEUE          | /fcrepo/activemq/queue          | fedora                            |             |
+| FCREPO_ACTIVEMQ_TOPIC          | /fcrepo/activemq/topic          | fedora                            |             |
+| FCREPO_BINARYSTORAGE_TYPE      | /fcrepo/binarystorage/type      | file                              |             |
+| FCREPO_CATALINA_OPTS           | /fcrepo/catalina/opts           |                                   |             |
+| FCREPO_DB_HOST                 | /fcrepo/db/host                 | mariadb                           |             |
+| FCREPO_DB_NAME                 | /fcrepo/db/name                 | fcrepo                            |             |
```
nigelgbanks, line 36: See notes above about environment variables.

fcrepo6/rootfs/etc/cont-init.d/03-fcrepo-setup.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/with-contenv bash
+set -e
+
+function execute_sql_file {
+    execute-sql-file.sh \
+        --driver "${FCREPO_PERSISTENCE_TYPE}" \
```
nigelgbanks, line 6: See notes above for fcrepo4, May or may not be relevant to this since `repository.json` is no longer in use.

fcrepo6/rootfs/etc/confd/templates/setenv.sh.tmpl
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+export JAVA_OPTS="{{ getv "/java/opts" "" }}"
+export CATALINA_OPTS="{{ getv "/catalina/opts" (getenv "FCREPO_CATALINA_OPTS") }}"
+export CATALINA_OPTS="${CATALINA_OPTS} -Dfcrepo.home=/data/home"
+export CATALINA_OPTS="${CATALINA_OPTS} -Dfcrepo.velocity.runtime.log=/opt/tomcat/logs/velocity.log"
+export CATALINA_OPTS="${CATALINA_OPTS} -Dfcrepo.jms.baseUrl=http://{{ index (lookupIP (getenv "HOSTNAME")) 0 }}/fcrepo/rest"
+export CATALINA_OPTS="${CATALINA_OPTS} -Dfcrepo.external.content.allowed=/opt/tomcat/conf/allowed-external-content.txt"
+export CATALINA_OPTS="${CATALINA_OPTS} -Dfcrepo.autoversioning.enabled=false"
+export CATALINA_OPTS="${CATALINA_OPTS} -Dfcrepo.db.url=jdbc:{{ getv "/persistence/type" (getenv "FCREPO_PERSISTENCE_TYPE") }}://{{ getv "/db/host" (getenv "FCREPO_DB_HOST") }}:{{ getv "/db/port" (getenv "FCREPO_DB_PORT") }}/fcrepo"
```
nigelgbanks, line 9: I assume there is no `repository.json` in fcrepo6 seems like the defaults here would cause problems? Like the default for `FCREPO_PERSISTENCE_TYPE` is file?

nigelgbanks (commented):
Lot had changed since the last time I gave a look so I thought I would look in again, it looks like it has come along way. I just mainly had some concerns around the environment variables and what looks like potential issues with defaults.

dannylamb:
Thanks for the feedback @nigelgbanks . I agree, the defaults are problematic for fcrepo6, but I wasn't sure how best to handle that, as well as splitting apart the mysql vs postgres variables.  I could've sworn I grepped for their usage and came up short, which is why I just straight up overwrote them, but I guess not.  One of those bad decisions made during "make it work" mode.

I'll make another pass at this and try to re-work things...  I"m :facepalm: 'ing hard just looking at those templates which need not be templates....

fcrepo/Dockerfile
```diff
@@ -27,18 +33,19 @@ ENV \
     FCREPO_ACTIVEMQ_QUEUE=fedora \
     FCREPO_ACTIVEMQ_TOPIC=fedora \
     FCREPO_BINARYSTORAGE_TYPE=file \
-    FCREPO_DB_MYSQL_HOST=mariadb \
-    FCREPO_DB_MYSQL_PORT=3306 \
```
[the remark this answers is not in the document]
noahwsmith: That may have been my problem- Fedora couldn't find its database when I was using secrets.

nigelgbanks:
You probably had them as templates before the rebase and changes to the confd system which makes sense.

Pulling the sandbox in as a demo makes a lot of sense, let me know if/when you want me to setup default_content in https://github.com/dannylamb/islandora-sandbox.

nigelgbanks:
@dannylamb I merged main into your branch to unblock @zacanbot and his work on Kubernetes.

alpaca/Dockerfile
```diff
@@ -29,7 +29,6 @@ RUN bin/start && \
 
 # Triple indexing
 RUN bin/start && \
-    bin/client -r 10 -d 5  "feature:install fcrepo-indexing-triplestore" && \
```
nigelgbanks, line 32: This service is keeping me from being able to run on Arm for some reason I cannot sort out (Karaf makes me wanna pull my hair out), is it not needed on main either?

dannylamb:
Ok, I've finally wrapped up all the code/test stuff for our php code and can circle back to this.  Thank you for your patience :bow: .  We'll get there.

drupal/rootfs/etc/islandora/utilities.sh
```diff
@@ -360,29 +359,31 @@ function configure_islandora_module {
...
+	drush -l "${site_url}" role:create fedoraadmin fedoraAdmin
+    fi
+    drush -l "${site_url}" -y user:role:add fedoraadmin admin
 }
 
 # After enabling and importing features a number of configurations need to be updated.
 function configure_islandora_default_module {
+    if ! drush pm-list --pipe --type=module --status=enabled --no-core | grep -q islandora_defaults; then
+	echo "islandora_defaults is not installed.  Skipping configuration"
+        return 0
```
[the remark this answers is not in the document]
dannylamb: nah, i'll fix 'em.  dang tabs...

dannylamb:
@nigelgbanks @noahwsmith I've made another pass.  Any :eyes: on this would be appreciated.

noahwsmith:
Hi Danny - I'm pretty deep underwater at the moment, but can test over the weekend if this is still open.

nigelgbanks (approved):

nigelgbanks:
Changes look good, and everything appears to work for me locally

PR #96, opened 2020-12-08
Refactor how environment variables secrets and defaults are set.

nigelgbanks:
Removes confd generated scripts
Makes clearer the order in which configuration can be provided
Secrets no longer need an accompanying environment variable.


nigelgbanks:
When merged this requires that https://github.com/Islandora-Devops/isle-dc/pull/133 is also merged.

nigelgbanks:
Looks like changing the base image and rebuilding the whole thing now exceeds the 30 minute time limit.

nigelgbanks:
This should wait until https://github.com/Islandora-Devops/isle-buildkit/pull/104 is merge as changes will be required. I can't seem to put the pull request into draft mode though so a text warning will have to suffice!

nigelgbanks:
@dannylamb I've sorted out the merge conflicts and tested Matomo this should now be ready for you to continue your review.

nigelgbanks:
Also a reminder that this should be tested with https://github.com/Islandora-Devops/isle-dc/pull/133 and that the handle pull request mentioned [here](https://github.com/Islandora-Devops/isle-buildkit/pull/96#issuecomment-759816425) will be handled after this is merged instead of before.

dannylamb (approved):
Tested with https://github.com/Islandora-Devops/isle-dc/pull/133 and it passes basic smoke/sanity tests.  The pattern introduced here is good. :rocket:

PR #97, opened 2020-12-08
Clear git repository cache and re-attempt if checkout for given commi…

nigelgbanks:
…t fails, clear cache on script failure.

Also ran a linter which changed the formatting somewhat.


dannylamb (approved):
Yeah, I've bumped into that issue quite a bit when developing.  Nice fix!

PR #105, opened 2021-01-13
Release versioning

dannylamb:
Follow up to https://github.com/Islandora-Devops/isle-buildkit/pull/99

With this PR
- Any commit to main updates the `latest` image
- Any version makes images tagged with the versions name/number
- Any commit that's pushed makes images tagged with the commit SHA

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -28,8 +28,8 @@ jobs:
         echo '{"experimental": "enabled"}' > ~/.docker/config.json
     - name: Set TAG Environment Variable
       run: |
-        export TAG=$(echo "${GITHUB_REF}" | cut -d "/" -f3)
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -1,11 +1,9 @@
 name: CI
 on:
-  pull_request:
-    branches:
-      - main
+  push:
```
nigelgbanks, line 3: Will this still trigger on pull requests after this change?

nigelgbanks: Also it would seem all branches will now push to docker hub, is that the intention?

nigelgbanks (approved):

.github/workflows/push.yml
nigelgbanks: Talked offline, and all commits that are pushed are to become images and for pull requests, we'll see if it works if not we'll do another change.

PR #110, opened 2021-02-11
Use explicit frontend version of buildkit, should help with some down…

nigelgbanks:
…stream bugs.

Background info: https://hub.docker.com/r/docker/dockerfile


nigelgbanks:
This change should be Docker version agnostic as the frontend is downloaded and used by Docker rather than bundled with it.

dannylamb:
Seems like a smart thing to do.  I'll give it a basic smoke test and report back.

dannylamb (approved):
Everything looks good :rocket:

PR #114, opened 2021-02-26
Temporary fix for Kubernetes

nigelgbanks:
Allows kuberentes users to start up the images.


nigelgbanks:
So small and a blocker for others that I just merged it without the normal workflow.

PR #118, opened 2021-03-07
Use the version 0.0.4 of the build plugin

nigelgbanks:
Drops support for non-buildkit builds. Adds the following:

- multi-arch builds (amd64, arm64, etc)
- custom build of confd to support arm64
- unique build caches per image (should reduce cache misses)
- fixes fits (mediainfo, exifdata, file info)
- update gradle version to latest (match plugin)
- child images of tomcat can override tomcat properties
- child images of nginx can override php/nginx properties
- updated to use latest alpine 3.13
- php is no 7.4
- imagemagick only builds what is required (no longer make clibs, etc)
- general refactors


nigelgbanks:
This should wait till https://github.com/Islandora-Devops/isle-buildkit/pull/95 is already in, I'll rebase and make changes to support fcrepo6, etc. after that point.

nigelgbanks:
Gonna close this as the pr action doesn't push to the remote this was just to check that that worked as well for both x86 and arm64

PR #121, opened 2021-03-16
Number of fixes / automation to address issues created from previous refactor

nigelgbanks:
**Summary:**
- confd settings are no longer forced to be prefixed by image
  - Allows images to effectively share variables like `JWT_PRIVATE_KEY`
  - Allows images to override settings like `DB_MYSQL_HOST` with image specific values `FCREPO_DB_MYSQL_HOST` or to default to `DB_MYSQL_HOST`.
- Adds automated testing framework that runs Docker images or uses Docker Compose to tests assumptions about behaviour.
- Containers now return an appropriate exit code depending on what error occurred (handles signals properly and propagates exit codes from failed services)
- Refactored some common behaviour into shell scripts `confd-import-environment.sh` and `confd-override-environment.sh`
- JWT confd templates, etc has moved into the base image where-in a new group `jwt` is required for the service user to access the files.
- Fixed issues found in testing (i.e. ETCD no longer worked)


nigelgbanks:
A lot of the changes are in the first commit https://github.com/Islandora-Devops/isle-buildkit/pull/121/commits/d81eaf5346b7526bb60ad295990b1c6422b7f797 which removes prefixes from the `confd` settings. The interesting bit are in the commits that follow that.

PR #133, opened 2021-04-01
Fixes for upgrade to latest Alpine

nigelgbanks:


dannylamb:
This is pretty uncontroversial.  Looks mostly like config file boilerplate from upstream.

I'm pulling in the images now from Dockerhub to test. :rocket:

dannylamb:
Well I have a drupal site so I guess it works :smile: 
image

PR #139, opened 2021-04-13
Fix illegal instruction issue in confd (arm64 only).

nigelgbanks:
[upx](https://upx.github.io/) the tool used to minimise the go executable (confd) does not play well with older Linux kernels, so we no longer minimise the executable. This is required to run on older Linux kernels with Arm64 architecture.


dannylamb (approved):
:+1:

Issue #141, opened 2021-04-14
islandora_bagger requires php7-intl


noahwsmith:
```
islandora_bagger noah$ composer install

Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for whikloj/bagittools 1.4.0 -> satisfiable by whikloj/bagittools[1.4.0].
    - whikloj/bagittools 1.4.0 requires ext-intl * -> the requested PHP extension intl is missing from your system.
```

I believe that should be added here: https://github.com/Islandora-Devops/isle-buildkit/blob/main/nginx/Dockerfile#L8

So I'll make a PR to add this for future releases, unless this dependency is going to go away. 

cc @mjordan @nigelgbanks @dannylamb @whikloj

noahwsmith:
Thanks!

PR #142, opened 2021-04-14
Adds php7-intl

noahwsmith:
Addresses #141. Just adding another PHP extension which is a dependency for Islandora Bagger, among other things.


noahwsmith:
Ugh, there are more requirements for that module. Hold up on accepting this- I'll add them to the same PR.

nigelgbanks (approved):

PR #149, opened 2021-06-28
Remove GPG checks as they occasionally fail to contact the keyserver,…

nigelgbanks:
… and we already verify the download by its checksum.


nigelgbanks:
This shouldn't require any testing as a successful build means the change worked.

dannylamb (approved):
:rocket: 

Looks like those spotty key servers are holding up our automated builds.  Good to see them go.

PR #150, opened 2021-06-28
Correcting newlines on template for allowed external content in fcrepo

dannylamb:
Adding `-` to force the template to make each allowed host on its own line.

RDF for files not stored in Fedora were failing to index due to needing both `http` and `https` domains allowed.


seth-shaw-unlv:
Looks like this might keep failing until #149 is merged. See https://github.com/Islandora-Devops/isle-buildkit/runs/2933265016?check_suite_focus=true#step:5:5762

nigelgbanks (approved):
Tested locally and works as described.

PR #153, opened 2021-07-09
Updated plugin / gradle to latest

nigelgbanks:
Can now perform build scans with `--scan` argument.

Ideally this should be merged first: https://github.com/Islandora-Devops/isle-gradle-docker-plugin/pull/11


nigelgbanks:
Not sure how the "circular" dependency has been introduced here, I can't reproduce it locally, so I'm gonna delete my entire cache, etc and see if I can reproduce it locally to debug further.

dannylamb (approved):
Build's are passing and this looks good :rocket:

Issue #154, opened 2021-07-19
Handle .txt files


kayakr:
If I add a Repository Item with Model=Binary, then add an associated Media of type File with a `text/plain` e.g. `test-text-file.txt` the media is created but the default media link to download the file https://islandora.traefik.me/_flysystem/fedora/2021-07/test-text-file.txt generates a 404 via nginx.

If visit the fcrepo record, e.g. http://islandora.traefik.me:8081/fcrepo/rest/2021-07/test-text-file.txt/fcr:metadata, and click 'Download Content' button the file is downloaded ok http://islandora.traefik.me:8081/fcrepo/rest/2021-07/test-text-file.txt

nigelgbanks:
Sounds like either a bug in Islandora or in the configuration of Drupal / Islandora. Moving the ticket to either https://github.com/Islandora/islandora/pulls or https://github.com/Islandora-Devops/isle-dc/issues, will increase the chances of someone tackling it.

kayakr:
@nigelgbanks I just tried this on an existing Drupal using Ansible Playbook and http://islandora8-demo.local:8000/_flysystem/fedora/2021-07/test-text-file.txt serves without error, so it looks like an isle-dc issue, or it's related to the nginx config at https://github.com/Islandora-Devops/isle-buildkit/blob/main/drupal/rootfs/etc/nginx/shared/drupal.defaults.conf

nigelgbanks:
My bad you are right, this is the config responsible: https://github.com/Islandora-Devops/isle-buildkit/blob/main/drupal/rootfs/etc/nginx/shared/drupal.defaults.conf#L21-L24 I thought Danny, had resolved this before, but maybe I've re-introduced it.

nigelgbanks:
Ahh this was Danny's fix: https://github.com/Islandora-Devops/isle-buildkit/blob/main/drupal/rootfs/etc/nginx/shared/drupal.defaults.conf#L15-L18 it only affects those txt files servied from the sites files directory, we also need to account for flystem it would seem.

nigelgbanks:
Odd though I would expect it to be 403 forbidden if nginx was denying the request

nigelgbanks:
Ya seems like the rule:

```conf
# Allow access to .txt files served by flysystem    
location ~ ^/_flysystem/.*\.txt$ {                  
    allow all;                                      
}
```
Does not resolve the issue.

nigelgbanks:
[...]

PR #155, opened 2021-07-20
Allow txt/log files to be served from flysystem.

nigelgbanks:
Addresses issue #154


kayakr:
@nigelgbanks I manually applied the same fix to an isle-dc Drupal container and `.txt` file is now served via flysystem. Thanks for the prompt fix.

PR #157, opened 2021-10-13
Remove file.stream references.

ruebot:
- Align with
  - https://github.com/Islandora/Alpaca/pull/78
  - https://github.com/Islandora/Alpaca/releases/tag/Alpaca-1.0.5
  - https://github.com/Islandora-Devops/isle-buildkit/blob/b3e3f46545e94a25ffe18613e63d2b984621cd7e/alpaca/Dockerfile#L8


nigelgbanks:
Hmm its passing for me locally on main and on your branch, I think the tests may time out over heavy load. Since the system makes very heavy use of caching it's hard to reproduce timing failures like this.  For the purpose of CI perhaps I should separate the testing into a different task and prevent them from all running in parallel.

nigelgbanks (approved):

nigelgbanks:
@ruebot sorry it took me so long to test this afterwards, kinda went off the deep end switching all my Linux machines to NixOS and learning that, took me a while to get a working system back in place.

PR #161, opened 2021-10-17
Make Pull Requests less likely to timeout on test

nigelgbanks:
Changes it so a build occurs before any tests are run, tests are also
run in sequence rather than parallel to ensure enough system resources
are available to prevent timeouts.


nigelgbanks:
Adds about 10 mins to build time

PR #163, opened 2021-11-22
Automatically update DockerHub image documentation.

nigelgbanks:
Adds Github action to update documentation in DockerHub when main branch is updated and a readme file has been modified.


nigelgbanks:
Merged as it doesn't affect any generated assets and we can't tested it until it's been merged into the main branch.

PR #165, opened 2021-11-22
Fix logging errors with cantaloupe due to find / replace bug

nigelgbanks:


scottdoane:
I pulled the PR, built it locally and ran it locally. Appears to be functional now.

PR #166, opened 2021-11-29
Adding max_input_vars to php.ini

dannylamb:
From a client site with a lot of fields (like 150+), forms for changing display and form modes started to have their ajax fail.  When trying to do something that triggers AJAX like hiding a field in a display mode,  you'll get an infinite spinning wheel.  This is b/c you'll bump up against the limit of how large form POSTs are allowed to get.

This PR fixes the issue by letting you configure the `max_input_vars` configuration for PHP and sets the default to something reasonably high so most people won't have to mess with this.


nigelgbanks (approved):

dannylamb:
:bow: thanks Nigel

PR #167, opened 2021-12-10
log4j zero day response

noahwsmith:
In response to https://www.lunasec.io/docs/blog/log4j-zero-day/

@DonRichards and I decided to shim the protection into all JAVA_OPTS and force a build.

### DIFF
--- a/cantaloupe/rootfs/opt/tomcat/bin/setenv.sh
+++ b/cantaloupe/rootfs/opt/tomcat/bin/setenv.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/with-contenv bash
-export JAVA_OPTS="${TOMCAT_JAVA_OPTS}"
+export JAVA_OPTS="${TOMCAT_JAVA_OPTS} ‐Dlog4j2.formatMsgNoLookups=True"
 export CATALINA_OPTS="${TOMCAT_CATALINA_OPTS}"
 export CATALINA_OPTS="${CATALINA_OPTS} -Dcantaloupe.config=/opt/tomcat/conf/cantaloupe.properties"
 export CATALINA_OPTS="${CATALINA_OPTS} -Dorg.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH=true"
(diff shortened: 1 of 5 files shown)


nigelgbanks:
Unfortunately when testing, I've encountered an issue:

```log
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] 00-container-environment-00-init.sh: executing... 
[cont-init.d] 00-container-environment-00-init.sh: exited 0.
[cont-init.d] 00-container-environment-01-override-tomcat.sh: executing... 
[cont-init.d] 00-container-environment-01-override-tomcat.sh: exited 0.
[cont-init.d] 00-container-environment-02-database-defaults.sh: executing... 
[cont-init.d] 00-container-environment-02-database-defaults.sh: exited 0.
[cont-init.d] 01-confd-render-templates.sh: executing... 
[cont-init.d] 01-confd-render-templates.sh: exited 0.
[cont-init.d] 02-ngnix-install.sh: executing... 
[cont-init.d] 02-ngnix-install.sh: exited 0.
[cont-init.d] done.
[services.d] starting services
[services.d] done.
Error: Could not find or load main class ‐Dlog4j2.formatMsgNoLookups=True
[services.d] service tomcat finish: executing...
[services.d] service tomcat exiting with exit code: 1
[services.d] service nginx finish: executing...
[services.d] service nginx exiting with exit code: 0
[cont-finish.d] executing container finish scripts...
[cont-finish.d] done.
[s6-finish] waiting for services.
[s6-finish] sending all processes the TERM signal.
[s6-finish] sending all processes the KILL signal and exiting.
```
I'll investigate it further

nigelgbanks:
So turns out only fits uses log4j and it's using a super old version `1.2.17` for which the workaround in the [CVE](https://nvd.nist.gov/vuln/detail/CVE-2021-44228) will not work since its `2.10`.  It would appear that this is of limited risk for users as fits isn't typically exposed so:

> An attacker who can control log messages or log message parameters can execute arbitrary code loaded from LDAP servers when message lookup substitution is enabled. 

Should not be applicable. That being said it would be nice to update but fits isn't exactly updated frequently https://github.com/harvard-lts/fits.

nigelgbanks:
I'm going to close this for now as it breaks the application, and a lower priority work can be done to investigate what can be done about fits.

nigelgbanks:
Ah I just noticed from Luke's comment on slack about solr, we're currently on and earlier version so again it has the same problem as fits in that this work around doesn't work for earlier versions of log4j.

PR #176, opened 2021-12-17
[issue-172] Use explicit commits in URLs such that

nigelgbanks:
Code can change when pulling from a master branch, which causes checksum
checks to fail.

This can changes master to instead refer to explicit commits hashes.

The output docker images is byte for byte the same though.


dannylamb:
Builds locally for me and seems to fix the build issues in our CI/CD.  :rocket:

dannylamb (approved):

PR #178, opened 2022-01-03
Fix config/sync permissions

DonRichards:
PLEASE TEST FIRST
I didn't develop this on a truly vanilla branch so please test this before merging. This should work fine but I just wanted to give a heads up. I normally test on a completely fresh build and that isn't the case for this pull request.

Simple fix to make the directory and set the permissions to the `/var/www/drupal/config/sync` directory. 

## To see error before fixing
Go to /admin/config/development/configuration/full/import and you'll see this error message.

Screenshot

## To check fix
Go to /admin/config/development/configuration/full/import and you'll see no error message.

Try changing a config and running a config export to verify. `docker-compose exec drupal bash -lc 'drush config:export -y'` and the result should look like this.

```shell
 // The .yml files in your export directory (/var/www/drupal/config/sync) will be deleted and replaced with the active
 // config.: yes.

 [success] Configuration successfully exported to /var/www/drupal/config/sync.
/var/www/drupal/config/sync
```
## Failure will look like this.
Screenshot failed

### DIFF
--- a/drupal/Dockerfile
+++ b/drupal/Dockerfile
@@ -29,6 +29,10 @@ RUN --mount=type=cache,id=drupal-apk,sharing=locked,from=cache,target=/var/cache
     addgroup nginx jwt && \
     cleanup.sh
 
+RUN mkdir -p /var/www/drupal/config/sync && \
+    chown -R nginx /var/www/drupal/config/sync && \
+    chmod -R 775 /var/www/drupal/config/sync
+
 ENV \
     DRUPAL_DEFAULT_ACCOUNT_EMAIL=webmaster@localhost.com \
     DRUPAL_DEFAULT_ACCOUNT_NAME=admin \


nigelgbanks:
@DonRichards your original issue likely arose from bind mounting `/var/www/drupal/config` into the container. Likely the files in the folder were owned by your host user as a result of some other process outside this project. All interactions with `drush` run through the script https://github.com/Islandora-Devops/isle-buildkit/blob/main/drupal/rootfs/usr/local/bin/drush#L9 which forces the users to be `nginx`.

For example, I've attached a minimal project that brings up a Drupal site:

[example.zip](https://github.com/Islandora-Devops/isle-buildkit/files/7863013/example.zip)

You can see the steps in `run.sh` after it is running you can export a config without the changes made here.

DonRichards:
@nigelgbanks Yeah but it doesn't work once in isle-dc. When spinning up isle-dc the problem persists. Putting the fix there seemed like a bandaid when the problem originates with the original image. I can see how it's supposed to work but I'm testing from a completely vanilla build. I wipe the docker image cache and refetch everything when testing.

DonRichards:
If I had to guess, there's an order of operations issue here

DonRichards:
This fix was submitted a month ago so it might be fixed in other ways by now. I'll test on my next build

nigelgbanks:
> Putting the fix there seemed like a bandaid when the problem originates with the original image.

I don't follow that the problem originates with the original image. 

> When spinning up isle-dc the problem persists. 

I just did a `make demo` on a clean checkout of the `development` branch of isle-dc and was not able to replicate your issue.

```bash
# docker-compose exec drupal ls -lah /var/www/drupal/config/sync 
drwxr-xr-x    2 nginx    nginx      68.0K Jan 13 23:10 .
```

If you can provide repeatable steps for your issue I could potentially find the source of the problem.

PR #179, opened 2022-01-10
Refactor how we set/restore ownership of drupal settings file to allow reuse...

jasonhildebrand:
...in containers based on this one.

The net effect of this PR on its own is none - everything works as before, I have simply factored out these two functions.

I'll be submitting a PR against isle-dc shortly where I make use of these functions to fix an issue occurring there.

### DIFF
--- a/drupal/rootfs/etc/islandora/utilities.sh
+++ b/drupal/rootfs/etc/islandora/utilities.sh
@@ -312,12 +312,14 @@ function restore_settings_ownership {
     local subdir=$(drupal_site_env "${site}" "SUBDIR")
     local site_directory=$(realpath "${drupal_root}/sites/${subdir}")
 
-    # Restore owner/group to previous value
+    # Restore owner/group to previous value.  
+    # When the codebase is bind-mounted, this ensures the file remains readable/writable by the host user.
     if [ ! -z "${previous_owner_group}" ]; then
-        #echo "restoring ${site_directory}/settings.php"
-        #echo "previous_owner_group ${previous_owner_group}"
         chown "${previous_owner_group}" "${site_directory}/settings.php"
     fi
+
+    # Restrict access to settings.php
+    chmod 444 "${site_directory}/settings.php"
 }
 
 
@@ -370,9 +372,6 @@ function update_settings_php {
 
     # Restore owner/group to previous value
     restore_settings_ownership ${site} ${previous_owner_group}
-
-    # Restrict access to settings.php
-    chmod 444 "${site_directory}/settings.php"
 }
 
 # Enable module and apply configuration.
--- a/drupal/rootfs/etc/islandora/utilities.sh
+++ b/drupal/rootfs/etc/islandora/utilities.sh
@@ -292,7 +292,7 @@ function allow_settings_modifications {
     local subdir=$(drupal_site_env "${site}" "SUBDIR")
     local site_directory=$(realpath "${drupal_root}/sites/${subdir}")
 
-    # send debug output stderr because call captures output from this function
+    # send debug output to stderr because the caller typically captures output from this function.
     #>&2 echo "adjusting ownership of ${site_directory}/settings.php"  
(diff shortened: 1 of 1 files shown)


jasonhildebrand:
Here is the issue this refactoring helps to fix: https://github.com/Islandora-Devops/isle-dc/pull/208

drupal/rootfs/etc/islandora/utilities.sh
```diff
@@ -336,9 +369,7 @@ function update_settings_php {
     fi
 
     # Restore owner/group to previous value
-    if [ ! -z "${previous_owner_group}" ]; then
-        chown "${previous_owner_group}" "${site_directory}/settings.php"
-    fi
+    restore_settings_ownership ${site} ${previous_owner_group}
 
     # Restrict access to settings.php
     chmod 444 "${site_directory}/settings.php"
```
nigelgbanks, line 375: We should move the correction of permissions into `restore_settings_ownership`
```bash
# Restrict access to settings.php
chmod 444 "${site_directory}/settings.php"
```

drupal/rootfs/etc/islandora/utilities.sh
```diff
@@ -284,6 +284,43 @@ function fedora_url {
...
+function allow_settings_modifications {
+    local site="${1}"; shift
+    local drupal_root=$(drush drupal:directory)
+    local subdir=$(drupal_site_env "${site}" "SUBDIR")
+    local site_directory=$(realpath "${drupal_root}/sites/${subdir}")
+
+    # send debug output to stderr because the caller typically captures output from this function.
+    #>&2 echo "adjusting ownership of ${site_directory}/settings.php"  
+    if [ -f "${site_directory}/settings.php" ]; then
+        previous_owner_group=$(stat -c "%u:%g" "${site_directory}/settings.php")
```
nigelgbanks, line 298: On review I don't know why we'd ever have an owner/group other than nginx:nginx (100:101)? With the possible exception if it's bind mounted. Possibly all files in `/var/www/drupal` have the hosts users group to allow for host edits or something... That being said it doesn't hurt anything. Just a thought as to why did we do this in the first place.

drupal/rootfs/etc/islandora/utilities.sh
```diff
@@ -284,6 +284,43 @@ function fedora_url {
...
+    local site="${1}"; shift
+    local previous_owner_group="${1}"; shift
+    local drupal_root=$(drush drupal:directory)
+    local subdir=$(drupal_site_env "${site}" "SUBDIR")
+    local site_directory=$(realpath "${drupal_root}/sites/${subdir}")
+
+    # Restore owner/group to previous value
+    if [ ! -z "${previous_owner_group}" ]; then
+        #echo "restoring ${site_directory}/settings.php"
+        #echo "previous_owner_group ${previous_owner_group}"
```
nigelgbanks, line 318: We can remove the commented out `#echo` statements.

nigelgbanks (changes requested):
Good stuff, just a small gotcha with the restoration of perms being outside of the restore function.

drupal/rootfs/etc/islandora/utilities.sh
jasonhildebrand: Good catch re: permissions.  I've moved this into the function.

Restoring ownership is only important when it's bind mounted.  It ensures the file remains readable/writable by the host user.  I've added a comment in the code to this effect.

drupal/rootfs/etc/islandora/utilities.sh
jasonhildebrand: Agreed - I've taken them out.

drupal/rootfs/etc/islandora/utilities.sh
jasonhildebrand: Yes - good catch. I've made this change.

nigelgbanks (approved):

Issue #180, opened 2022-01-24
Add php7-ldap module to nginx image.


nigelgbanks:
Also have a think about documenting how to enable/disable `php` extensions. Should we run with more or less everything disabled then for each `php` extension we allow them to be enabled via some mechanism?

PR #181, opened 2022-01-24
Added LDAP extension to nginx/php image.

nigelgbanks:
Allows downstream use of ldap php extension.

Address https://github.com/Islandora-Devops/isle-buildkit/issues/180


ajstanley (approved):
This is about as straightforward as it gets.  Thanks for hitting it so quickly.
Everything looks good.

PR #185, opened 2022-02-08
Timeouts

dannylamb:
Here is every possible timeout you might wanna set if you're running into weird errors with really really big files.  This won't change anything for existing installs, as all defaults are still the same.  But this way you can add them to your docker-compose.yml and play with them if you need it.

At born digital we've used this to get a couple of clients out of a bind with ingests of very large files.


nginx/Dockerfile
```diff
@@ -39,9 +39,18 @@ RUN --mount=type=cache,id=nginx-apk,sharing=locked,from=cache,target=/var/cache/
...
 
 ENV \
+    NGINX_CLIENT_BODY_TIMEOUT=60 \
     NGINX_CLIENT_MAX_BODY_SIZE=0 \
     NGINX_ERROR_LOG_LEVEL=warn \
+    NGINX_FASTCGI_READ_TIMEOUT=60 \
+    NGINX_FASTCGI_SEND_TIMEOUT=60 \
+    NGINX_FASTCGI_CONNECT_TIMEOUT=60 \
     NGINX_KEEPALIVE_TIMEOUT=65 \
+    NGINX_LINGERING_TIMEOUT=60 \
```
nigelgbanks, line 49: The default is actual 5 seconds: https://nginx.org/en/docs/http/ngx_http_core_module.html#lingering_timeout

nginx/Dockerfile
```diff
@@ -53,6 +62,8 @@ ENV \
     PHP_MAX_INPUT_VARS=3000 \
     PHP_MEMORY_LIMIT=256M \
     PHP_POST_MAX_SIZE=128M \
+    PHP_PROCESS_CONTROL_TIMEOUT=60 \
+    PHP_REQUEST_TERMINATE_TIMEOUT=60 \
```
nigelgbanks, line 66: The default is `0` according to the documentation above for both of the `php` settings?

nginx/Dockerfile
```diff
@@ -39,9 +39,18 @@ RUN --mount=type=cache,id=nginx-apk,sharing=locked,from=cache,target=/var/cache/
     cleanup.sh
 
 ENV \
+    NGINX_CLIENT_BODY_TIMEOUT=60 \
     NGINX_CLIENT_MAX_BODY_SIZE=0 \
     NGINX_ERROR_LOG_LEVEL=warn \
+    NGINX_FASTCGI_READ_TIMEOUT=60 \
+    NGINX_FASTCGI_SEND_TIMEOUT=60 \
+    NGINX_FASTCGI_CONNECT_TIMEOUT=60 \
     NGINX_KEEPALIVE_TIMEOUT=65 \
```
nigelgbanks, line 48: Just noticed this isn't the same as the default: https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout

tomcat/Dockerfile
```diff
@@ -29,6 +29,20 @@ RUN --mount=type=cache,id=tomcat-apk,sharing=locked,from=cache,target=/var/cache
...
+    NGINX_FASTCGI_CONNECT_TIMEOUT=60 \
+    NGINX_KEEPALIVE_TIMEOUT=65 \
+    NGINX_LINGERING_TIMEOUT=60 \
+    NGINX_PROXY_READ_TIMEOUT=60 \
+    NGINX_PROXY_SEND_TIMEOUT=60 \
+    NGINX_PROXY_CONNECT_TIMEOUT=60 \
+    NGINX_SEND_TIMEOUT=60 \
+    NGINX_WORKER_CONNECTIONS=1024 \
+    NGINX_WORKER_PROCESSES=auto \
     TOMCAT_ADMIN_NAME=admin \
```
nigelgbanks, line 46: `NGINX_WORKER_CONNECTIONS` and `NGINX_WORKER_PROCESSES` are not used.

demo/Dockerfile
```diff
@@ -58,6 +58,6 @@ RUN --mount=type=cache,id=demo-composer,sharing=locked,target=/root/.composer/ca
     mkdir -p /var/www/drupal/web/sites/default/files/library-definitions && \
     cp /var/www/drupal/web/modules/contrib/openseadragon/openseadragon.json /var/www/drupal/web/sites/default/files/library-definitions
 
-FROM ${repository}/drupal:${tag}
+FROM ${repository}/drupal:${tag} AS drupal
```
nigelgbanks, line 61: This doesn't do anything as far as I can tell, not sure why it's included?

nigelgbanks (changes requested):
I've raised a pull request (https://github.com/Islandora-Devops/isle-buildkit/pull/188) against yours that moves all the configuration into `nginx.conf`  so its only managed in a single place. In reviewing your changes I've noticed that the reverse proxy for tomcat is no longer needed. Could you please review the other outstanding pull requests that are this repo?

dannylamb:
Oh nice, thanks @nigelgbanks !  I have a follow up PR for this with an extra tomcat timeout I missed.

Also big thanks for getting rid of the extra reverse proxy layer!

PR #196, opened 2022-03-04
Update default 3-store URL now that we've removed the nginx layer

dannylamb:
Just found this after poking around.  I fixed it in my local docker-compose.yml, but this'll stop everyone else from having to do it :smiley_cat:


nigelgbanks (approved):

nigelgbanks:
Good catch

Issue #197, opened 2022-03-16
Cache issue with the latest tag in the code-server image in versions after 1.0.0-alpha-12 


nigelgbanks:
I think this is arising from building a bad case and then relying on the cache going forward of that bad release. At the moment we're not using a lock file or version for code-server when doing the calls to `yarn` to install it (the assumption is it would always pull the latest) but since we're not specifying the version in the docker file the version is tied to the last time the build was run and if it got cached or not.

nigelgbanks:
@ysuarez yup

PR #198, opened 2022-03-16
Use lock file for installation of code-server

nigelgbanks:
For issue https://github.com/Islandora-Devops/isle-buildkit/issues/197


nigelgbanks:
Doing some testing locally, if goes well I'll also change the composer packages to use a lock file as well.

nigelgbanks:
This is ready to go and can be tested with the following:

```bash
docker run -it -p 8443:8443 -e CODE_SERVER_AUTHENTICATION=none islandora/code-server:7e9353c15fe2ad6c96a0d098270082a9509f87be
```

If you can then visit the IDE in the browser at [http://localhost:8443](http://localhost:8443), then it's working as intended.

jefferya (approved):
On WSL 2 Unbuntu, container built and was able to load in the browser
```
docker run -it -p 8443:8443 -e CODE_SERVER_AUTHENTICATION=none islandora/code-server:77a0730aef1fc2deb4a196b6c300dae2d639e5aa
```

PR #200, opened 2022-03-23
Bump timeout to 360 as a full cross platform rebuild with tests takes…

nigelgbanks:
… more than 60 minutes now.

Only `timeout-minutes` has changed the rest is formatting.


nigelgbanks:
Oh there isn't really anything needing testing here this just bumps the max timeout time for Github actions.

PR #228, opened 2022-06-17
Nginx drupal rule for protect files and directories was missing leadi…

nigelgbanks:
…ng slash for path.


rbos:
Confirmed working after updating to 1.0.2 upstream.

PR #241, opened 2022-09-06
Issue 224 Fixed - Dead code removed

g7morris:


hypercube/Dockerfile
```diff
@@ -29,9 +29,7 @@ RUN --mount=type=cache,id=hypercube-apk,sharing=locked,from=cache,target=/var/ca
     addgroup nginx jwt && \
     cleanup.sh
 
-ENV \
-    HYPERCUBE_FCREPO_URL=fcrepo/fcrepo/rest \
-    HYPERCUBE_LOG_LEVEL=info
+ENV HYPERCUBE_LOG_LEVEL=info
```
nigelgbanks, line 32: Still seems to be in use: https://github.com/Islandora-Devops/isle-buildkit/blob/main/hypercube/rootfs/etc/confd/templates/config.yaml.tmpl#L7

milliner/README.md
```diff
@@ -13,7 +13,7 @@ additional settings, volumes, ports, etc.
 | Environment Variable | Confd Key            | Default            | Description                                                                                       |
 | :------------------- | :------------------- | :----------------- | :------------------------------------------------------------------------------------------------ |
 | MILLINER_DRUPAL_URL  | /milliner/drupal/url | drupal:80          | Drupal URL                                                                                        |
-| MILLINER_FCREPO_URL  | /milliner/fcrepo/url | fcrepo/fcrepo/rest | Fcrepo Rest API URL                                                                               |
+| MILLINER_FCREPO_URL  | /milliner/fcrepo/url | fcrepo:8080/fcrepo/rest | Fcrepo Rest API URL                                                                          |
```
nigelgbanks, line 16: Just noticed now, but the URL variables for the microservices in the README.md files are missing the prefixed `http://`, so this should probably be fixed across all microservices.

nigelgbanks: Also, can you make sure the alignment of the tables are kept? This can be done with <https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one> if using VS Code, and I'm sure other auto-formatters exist for other IDEs.

hypercube/Dockerfile
nigelgbanks: If this isn't needed, it should be removed from the `config.yaml.tmpl` file as well.

nigelgbanks (changes requested):
Just a few things remaining to be cleaned up.

milliner/README.md
g7morris: Roger that! Will do.

hypercube/Dockerfile
g7morris: Apologies for delay. Removed now

milliner/README.md
g7morris: Not sure if I did this right in terms of alignment. I do use that auto-formatter but it seemed a bit thick on moving things around properly. Or perhaps it's operator was? 😄

g7morris: Also tested this on my local. no issues discovered

nigelgbanks (approved):

hypercube/Dockerfile
nigelgbanks: lol, turns out it was supposed to be there. https://islandora.slack.com/archives/CM6F4C4VA/p1668431519198479

Issue #243, opened 2022-09-07
Upgrade Solr


nigelgbanks:
Currently, Solr is using arguments for `ps` that are not supported by busybox (therefore also Alpine).

```
idc-solr-1        | [cont-init.d] 02-setup-environment.sh: exited 0.
idc-solr-1        | [cont-init.d] 02-solr-setup.sh: executing... 
idc-solr-1        | [cont-init.d] 02-solr-setup.sh: exited 0.
idc-solr-1        | [cont-init.d] done.
idc-solr-1        | [services.d] starting services
idc-solr-1        | [services.d] done.
idc-solr-1        | ps: unrecognized option: p
idc-solr-1        | BusyBox v1.33.1 () multi-call binary.
idc-solr-1        | 
idc-solr-1        | Usage: ps [-o COL1,COL2=HEADER] [-T]
idc-solr-1        | 
idc-solr-1        | Show list of processes
idc-solr-1        | 
idc-solr-1        | 	-o COL1,COL2=HEADER	Select columns for display
idc-solr-1        | 	-T			Show threads
idc-solr-1        | This script relies on a version of ps that supports the -p flag.
idc-solr-1        | 
idc-solr-1        | Please install a POSIX compliant version and try again.
idc-solr-1        | [cont-finish.d] executing container finish scripts...
idc-solr-1        | [cont-finish.d] done.
idc-solr-1        | [s6-finish] waiting for services.
idc-solr-1        | [s6-finish] sending all processes the TERM signal.
idc-solr-1        | [s6-finish] sending all processes the KILL signal and exiting.
idc-solr-1 exited with code 0
```

This has been resolved [here](https://issues.apache.org/jira/browse/SOLR-16191).

Upgrading to versions 8.11.2 and 9.1 or greater should resolve the issue.

~~It is a low-priority issue, as only a warning is printed. Since the process always starts from a clean slate, no existing process for `solr` should be present.~~

Doesn't seem to be an issue on our older version of `solr` presently, that being said, `solr` should be updated regardless.

nigelgbanks:
This should wait until after https://github.com/Islandora-Devops/isle-buildkit/issues/215, as upgrading Solr will constitute a major release, as folks will need to update their Drupal site as well.

DonRichards:
@nigelgbanks Is this not the issue causing my exit code?

nigelgbanks:
[...]

PR #244, opened 2022-09-07
Solr 8.11.2 update

DonRichards:
This is untested.

I think this would work but I'm still testing it.

### DIFF
--- a/solr/Dockerfile
+++ b/solr/Dockerfile
@@ -3,8 +3,8 @@ ARG repository=local
 ARG tag=latest
 FROM --platform=$BUILDPLATFORM ${repository}/download:${tag} AS download
 
-ENV SOLR_VERSION="8.11.2"
-ENV SOLR_FILE_SHA256="54d6ebd392942f0798a60d50a910e26794b2c344ee97c2d9b50e678a7066d3a6"
+ARG SOLR_VERSION="8.11.2"
+ARG SOLR_FILE_SHA256="54d6ebd392942f0798a60d50a910e26794b2c344ee97c2d9b50e678a7066d3a6"
     
 RUN --mount=type=cache,id=solr-downloads,sharing=locked,target=/opt/downloads \
     SOLR_FILE="solr-${SOLR_VERSION}.tgz" && \
--- a/solr/Dockerfile
+++ b/solr/Dockerfile
@@ -3,9 +3,9 @@ ARG repository=local
 ARG tag=latest
 FROM --platform=$BUILDPLATFORM ${repository}/download:${tag} AS download
 
-ARG SOLR_VERSION="7.1.0"
-ARG SOLR_FILE_SHA256="5cd25cc2634e47efbb529658d6ddd406a7cd1b211affa26563a28db2d80b8133"
-
+ENV SOLR_VERSION="8.11.2"
+ENV SOLR_FILE_SHA256="54d6ebd392942f0798a60d50a910e26794b2c344ee97c2d9b50e678a7066d3a6"
+    
 RUN --mount=type=cache,id=solr-downloads,sharing=locked,target=/opt/downloads \
     SOLR_FILE="solr-${SOLR_VERSION}.tgz" && \
     SOLR_URL="https://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/${SOLR_FILE}" && \
--- a/solr/Dockerfile
+++ b/solr/Dockerfile
@@ -1,11 +1,12 @@
 # syntax=docker/dockerfile:1.2.1
 ARG repository=local
 ARG tag=latest
+ARG alpine=3.15.0
 FROM --platform=$BUILDPLATFORM ${repository}/download:${tag} AS download
 
 ARG SOLR_VERSION="8.11.2"
 ARG SOLR_FILE_SHA256="54d6ebd392942f0798a60d50a910e26794b2c344ee97c2d9b50e678a7066d3a6"
-    
+
 RUN --mount=type=cache,id=solr-downloads,sharing=locked,target=/opt/downloads \
     SOLR_FILE="solr-${SOLR_VERSION}.tgz" && \
     SOLR_URL="https://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/${SOLR_FILE}" && \
@@ -15,6 +16,7 @@ RUN --mount=type=cache,id=solr-downloads,sharing=locked,target=/opt/downloads \
         --file "${DOWNLOAD_CACHE_DIRECTORY}/${SOLR_FILE}" \
         docs example licenses server/solr/configsets
 
+FROM alpine:${alpine} AS cache
 FROM ${repository}/java:${tag}
 
 EXPOSE 8983
@@ -30,6 +32,10 @@ COPY --from=download /etc/group /etc/group
 COPY --from=download /etc/passwd /etc/passwd
 COPY --from=download /etc/shadow /etc/shadow
 COPY --from=download --chown=solr:solr /opt/solr /opt/solr
+
+# Solr requires GNU ps.
+RUN --mount=type=cache,id=solr-apk,sharing=locked,from=cache,target=/var/cache/apk \
+    apk add procps
 
 COPY rootfs /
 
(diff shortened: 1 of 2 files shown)


solr/README.md
```diff
@@ -42,3 +42,30 @@ additional settings, volumes, ports, etc.
...
 [Solr Logging]: https://lucene.apache.org/solr/guide/7_1/configuring-logging.html
 [Solr]: https://lucene.apache.org/solr/
+
+## Changing versions
+There is 3 values you need to update/change the version. 
+1. Solr version: found at [archive.apache.org](https://archive.apache.org/dist/lucene/solr)
+1. SOLR_KEYS: Generated using GPG and the acs file
+1. SOLR_FILE_SHA256: sha256sum of the tgz file
+
+```dockerfile
```
nigelgbanks, line 52: Same as above.

solr/Dockerfile
```diff
@@ -3,11 +3,13 @@ ARG repository=local
 ARG tag=latest
 FROM --platform=$BUILDPLATFORM ${repository}/download:${tag} AS download
 
+ENV SOLR_VERSION="8.11.2"
+ENV SOLR_KEYS="86EDB9C33B8517228E88A8F93E48C0C6EF362B9E"
```
nigelgbanks, line 7: `SOLR_KEYS` is not used anywhere here?

solr/README.md
```diff
@@ -42,3 +42,30 @@ additional settings, volumes, ports, etc.
 [Solr Documentation]: https://lucene.apache.org/solr/guide/7_1/
 [Solr Logging]: https://lucene.apache.org/solr/guide/7_1/configuring-logging.html
 [Solr]: https://lucene.apache.org/solr/
+
+## Changing versions
+There is 3 values you need to update/change the version. 
+1. Solr version: found at [archive.apache.org](https://archive.apache.org/dist/lucene/solr)
+1. SOLR_KEYS: Generated using GPG and the acs file
```
nigelgbanks, line 49: Same as above.

solr/README.md
```diff
@@ -42,3 +42,30 @@ additional settings, volumes, ports, etc.
 [Solr Documentation]: https://lucene.apache.org/solr/guide/7_1/
 [Solr Logging]: https://lucene.apache.org/solr/guide/7_1/configuring-logging.html
 [Solr]: https://lucene.apache.org/solr/
+
+## Changing versions
```
nigelgbanks, line 46: Please add a space after the header for markdown standards.

nigelgbanks (changes requested):
I have a few questions that might affect the upcoming release, and would require us to include some more info in the release notes, or do a minor/major bump.

1. Are any changes required on the Drupal side after this version bump?
2. Will this have an effect on someone's existing index (e.g. their `solr-data` Docker volume) if they were to switch images.
3.  Would they have to rebuild the index if it does?

Thanks!

solr/Dockerfile
DonRichards: No, this is an attempt to make it easier/consistent for developers to update without needing to modify strings embedded within logic. By consistent I'm assuming this is the right way to approach this and modify other Dockerfiles in the future to match.

solr/README.md
DonRichards: A trailing space character after "## Changing versions^"?

nigelgbanks: Sorry, newline after the header.

solr/Dockerfile
```diff
@@ -3,11 +3,13 @@ ARG repository=local
 ARG tag=latest
 FROM --platform=$BUILDPLATFORM ${repository}/download:${tag} AS download
 
+ENV SOLR_VERSION="8.11.2"
```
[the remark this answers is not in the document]
DonRichards: Any suggestions for a replacement? Should I switch it to **ARG**?

DonRichards:
@nigelgbanks To answer your original questions
1. Are any changes required on the Drupal side after this version bump?
> Not sure. It doesn't work for me yet so I imagine it might.

2. Will this have an effect on someone's existing index (e.g. their solr-data Docker volume) if they were to switch images.
> I'm glad you asked. I just looked it up [here](https://solr.apache.org/guide/8_0/major-changes-in-solr-8.html#rolling-upgrades-with-solr-8) and yes there is. Not sure how to integrate this into the codebase 

3. Would they have to rebuild the index if it does?
> They suggest always reindexing after an upgrade, I would assume a rebuild would be needed.

With all of that said, I'm not sure what the next steps should be for this PR.

solr/Dockerfile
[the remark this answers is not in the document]
nigelgbanks: `ARG` seems fine, we can change the other images for consistency as part of another issue (I'll create one).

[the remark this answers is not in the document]
nigelgbanks: https://github.com/Islandora-Devops/isle-buildkit/issues/245

solr/Dockerfile
nigelgbanks: If it's not being used, let's just remove it. As it's an extra step that has no effect.

solr/Dockerfile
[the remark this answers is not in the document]
DonRichards: OK, great. Thanks for adding the ticket. Changes made.

nigelgbanks:
> With all of that said, I'm not sure what the next steps should be for this PR.

At the least, we'll have to include it in the release notes. For `isle-dc` I'm not certain, there is a good solution as people can / are encouraged to maintain the `.env` file outside of source control, and we shouldn't be triggering re-builds of indexes if it's not appropriate.

DonRichards:
Sorry, I just realized I was clicking "Resolved" on your comments as though they were something for me to "resolve". lol

nigelgbanks:
> Sorry, I just realized I was clicking "Resolved" on your comments as though they were something for me to "resolve". lol

Haha no worries, after you have pushed the fix you can mark them as resolved.

solr/Dockerfile
nigelgbanks: The documentation in the README.md should also get updated to not reference `SOLR_KEYS` after it's removed.

DonRichards:
OK, I think I made all of the suggested changes. But to recap, on the release we should comment on how there is documentation from Solr on how to upgrade the index when upgrading Solr from 7 to 8. We should give an example command that the community has tested to upgrade Solr's index like the example they gave but adjusted for isle-dc

`./bin/solr start -c -Dsolr.http1=true -z localhost:2481/solr -s /path/to/solr/home`

They also had one last comment that we should probably spell out how to do it manually
> When all nodes have been upgraded to 8.0, restart each one without the -Dsolr.http1 parameter.

solr/Dockerfile
DonRichards: OK, done

DonRichards:
@nigelgbanks Do you see what's making the tests fail? I don't see what I need to change to get it to pass.

nigelgbanks:
Looks to be the same `ps` [issue](https://github.com/Islandora-Devops/isle-buildkit/issues/243):

```bash
BusyBox v1.34.1 (2021-11-23 00:57:35 UTC) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER] [-T]

Show list of processes

        -o COL1,COL2=HEADER     Select columns for display
        -T                      Show threads
This script relies on a version of ps that supports the -p flag.

Please install a POSIX compliant version and try again.
```

https://issues.apache.org/jira/browse/SOLR-16191

nigelgbanks:
Looks like I read the ticket wrong, on reviewing their actual changes made for 8.11.2 it seems like they now have a hard requirement on using GNU version of `ps` rather than supporting both.

nigelgbanks (approved):

nigelgbanks:
@DonRichards I've changed this to a draft just to prevent it from going in before we cut a minor release (which should be pretty soon). Subsequently, it should be good to go into the next major release, for which we'll document the need for folks to rebuild their `solr` index.

DonRichards:
@nigelgbanks are you good with this dropping the "draft" status now?

nigelgbanks:
@DonRichards whoops yea we'll start working on the 2.0.0 release.

nigelgbanks:
@DonRichards I've raised a ticket, so I don't forget about updating the release notes.

PR #257, opened 2022-10-28
Draft - Add m1/m2 processors for release builds

DonRichards:
I'm not sure if my local tests reflect how it would work on deployment. I would suggest thoroughly testing this.

This should add docker containers with M1/M2 support.

https://github.com/marketplace/actions/docker-setup-qemu

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Build/Test Docker images
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build test grype -PisCI=true --add-cpes-if-none --info
+          arguments: build test grype -PisCI=true --info
       - uses: actions/upload-artifact@v2
         with:
           name: Grype Reports
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -6,10 +6,7 @@ on:
 jobs:
   build:
     name: Build and Test PR
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     timeout-minutes: 360
     steps:
       - name: Checkout
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Build/Test Docker images
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: build test grype -PisCI=true --info
+          arguments: build test grype -PisCI=true --add-cpes-if-none --info
       - uses: actions/upload-artifact@v2
         with:
           name: Grype Reports
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -6,7 +6,10 @@ on:
 jobs:
   build:
     name: Build and Test PR
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 360
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


nigelgbanks:
The builds already produce docker containers that work with M1/M2 Macs, and have done since pretty much the begining of this repo. I've got an M1 and use the images produced by the build.

nigelgbanks:
I see you've added `linux/arm/v7` which is 32bit arm processor I don't think the executables used in the images will run on 32 bit Arm processors.

DonRichards:
@nigelgbanks I have an M1 and can't seem to get Isle to build. It builds fine on my Ubuntu machine and my non-M1 mac. I assumed it was an M1 issue. I had found a few places that suggested linux/arm/v7 | linux/arm/v8 was needed to officially support M1 chips.

nigelgbanks:
That's odd, I am able to build locally on my M1, can you add the log here?

```bash
./gradlew build --info 2>&1 | tee ~/Downloads/build.txt 
```

DonRichards:
@nigelgbanks  
[build.txt](https://github.com/Islandora-Devops/isle-buildkit/files/9903386/build.txt)

nigelgbanks:
```txt
#5 [internal] load metadata for docker.io/local/download:latest
#5 ERROR: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
------
 > [internal] load metadata for docker.io/local/download:latest:
------
ERROR: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: rpc error: code = Unknown desc = pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
:composer:build (Thread[#105,Execution worker Thread 6,5,main]) completed. Took 1.848 secs.
```

Which is a bit odd cause other images in the log that use the download image pull it just fine, like `base`.

I think the safest way for you forward would be to re-install Docker and try again. Something about that doesn't seem right.

DonRichards:
Are you using Docker Desktop on your M1?

nigelgbanks:
Yup

DonRichards:
A little over a week ago I reinstalled Docker Desktop just in case I installed something wrong. I can run other docker images without issue.

nigelgbanks:
Running and building are two different things, you should be able to run the images produced by GitHub actions with isle-dc? Tag '1.0.8' for example. You are successfully building most of the images it's just the composer one in the build log that failed. The reason it gave would make sense if it was pulling from a remote repository and failed to authenticate. Though it's the local repository in this case so weird, also the bit that fails for composer here is working for other images in your build log, so again weird. I can't really diagnose further from here maybe something is off in your build cache? Or with the installation, beyond that it's a black box at this point.

DonRichards:
@nigelgbanks If you run `make clean` then `make up` your's comes right up with isle-dc? Are you using a different make command to build, or are you using your local containers or something? Just trying to see what's different. One of our new employees just tried it and the build failed for his machine as well. I think there might be something missing from that is a prereq for this to work that I'm missing. 

I can get it to build without issue using Rosie's [portainer instructions](https://www.youtube.com/watch?v=dfc7WUGAmow). This tells me the buildkit isn't the issue at all, and I should close this PR. But I'm still left with a broken build for an M1 chip when trying to run the basic `make up` command.

nigelgbanks:
@DonRichards Might be a problem with isle-dc? The `portainer` example is using the same images, so seems more likely the `isle-dc` `make up` command is doing something funky? I've not used `isle-dc` for a year or so, but I updated to the latest `development` branch and selected `Generate Secrets` when prompted it takes a really, really, really long time to start, I get prompted for my root password several times, and it took like 45 minutes to start. I'm not certain, why it took so long, when I roll my own docker-compose file it takes a little under a minute to bring up a new site.

nigelgbanks:
@DonRichards capture your long from `make up` and raise an issue on the `isle-dc` repo, the log will likely tell the problem, it might be trying to use an executable on your host machine that you have not installed and that might be why it fails.

PR #258, opened 2022-11-01
Enable solr config even when Defaults is not present.

rosiel:
Ticket: https://github.com/Islandora-Devops/isle-dc/issues/305

Currently the ISLE-DC build does not configure solr when the Islandora Starter Site is used. I tracked it down to this function, `configure_islandora_default_module`.

The only thing that the function did was to set the Solr URL and port. 

I first just removed the whole `if` statement, but then decided it would be prudent to at least make sure `search_api` is enabled.

I did not rename the function because I am not sure how to manage the integration between versions of Buildkit and ISLE-DC.  Should we do that?

Tagging @aoelschlager

### DIFF
--- a/drupal/rootfs/etc/islandora/utilities.sh
+++ b/drupal/rootfs/etc/islandora/utilities.sh
@@ -427,13 +427,12 @@ function configure_islandora_module {
     drush -l "${site_url}" -y user:role:add fedoraadmin admin
 }
 
-# After enabling and importing features a number of configurations need to be updated.
+# Configure Solr port and host.
 function configure_islandora_default_module {
-    if ! drush pm-list --pipe --type=module --status=enabled --no-core | grep -q islandora_defaults; then
-        echo "islandora_defaults is not installed.  Skipping configuration"
+    if ! drush pm-list --pipe --type=module --status=enabled --no-core | grep -q search_api; then
+        echo "Search API is not installed.  Skipping configuration"
         return 0
     fi
-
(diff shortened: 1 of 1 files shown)


nigelgbanks:
Looks good to me.

> I did not rename the function because I am not sure how to manage the integration between versions of Buildkit and ISLE-DC. Should we do that?

I think you made the right choice, we do not know who out there in the world might be calling this function, so to prevent breaking users code the first step would be to deprecated it and rename/remove it on another release. 

That being said, I'm not a big fan of these utilities scripts but instead prefer folks used `composers` append feature to add things to `settings.php` like how sandbox does it:

- Reads variables from the environment to set Drupal configuration [here](https://github.com/Islandora-Devops/sandbox/blob/db5e3a613260ae0c002f4bb5c1c9a87b8f60d9e4/docker/sandbox-drupal/rootfs/var/www/drupal/assets/default.settings.php#L40-L50)
- Composer appends this file [here](https://github.com/Islandora-Devops/sandbox/blob/db5e3a613260ae0c002f4bb5c1c9a87b8f60d9e4/docker/sandbox-drupal/rootfs/var/www/drupal/composer.json#L121-L126) when installed.

I'll raise a ticket to deprecate and remove the utility scripts and add documentation on how users can use the environment variables to configure their images.

nigelgbanks:
I've raised https://github.com/Islandora-Devops/isle-buildkit/issues/260 to deal with the future deprecations / documentation.

PR #266, opened 2022-11-18
Update default fcrepo URL for Hypercube to use port 8080

dannylamb:
This resolves [issues/2190](https://github.com/Islandora/documentation/issues/2190). 

I guess we missed this default when removing the nginx proxy from the fedora container?  :shrug: Not sure why this didn't manifest sooner, but it's easily fixable.

### DIFF
--- a/hypercube/Dockerfile
+++ b/hypercube/Dockerfile
@@ -29,7 +29,7 @@ RUN --mount=type=cache,id=hypercube-apk,sharing=locked,from=cache,target=/var/ca
     addgroup nginx jwt && \
     cleanup.sh
 
-ENV HYPERCUBE_FCREPO_URL=fcrepo/fcrepo/rest \
+ENV HYPERCUBE_FCREPO_URL=fcrepo:8008/fcrepo/rest \
     HYPERCUBE_LOG_LEVEL=info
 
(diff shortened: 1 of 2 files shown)


hypercube/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN --mount=type=cache,id=hypercube-apk,sharing=locked,from=cache,target=/var/ca
     addgroup nginx jwt && \
     cleanup.sh
 
-ENV HYPERCUBE_FCREPO_URL=fcrepo/fcrepo/rest \
+ENV HYPERCUBE_FCREPO_URL=fcrepo:8008/fcrepo/rest \
```
ruebot, line 32: 8008 or 8080?

dannylamb: :facepalm:

dannylamb: I've taken the liberty to straight push that fix: https://github.com/Islandora-Devops/isle-buildkit/commit/b3e273a78ddbb7b1d8b052399299380372a2ec02

Sorry for the derp.

nigelgbanks: Wouldn't break it either way, since it's not used for anything other than the configuration parser expects some value to be present. Nonetheless best that it's correct :)

Issue #274, opened 2023-03-13
Matomo v4.13.0 missing images within containers created by #238


jefferya:
When using the Matomo container created by #238, I'm experiencing a few missing images

<img width="428" alt="Screenshot 2023-03-13 165842" src="https://user-images.githubusercontent.com/879785/224850972-1e12cb37-5c57-4301-8e7b-e389340e7c6e.png">

A Matomo issue [Morpheous icons missing on dashboard #16553](https://github.com/matomo-org/matomo/issues/16553) indicates when using the GitHub release,  the resulting download maybe missing any Git Submodules within the repository. The recommendation in the GitHub issue is to "Please use the releases published on https://builds.matomo.org/".

The current Matomo Dockerfile uses the GitHub release approach and misses icons included within the `plugins/Morpheus/icons/dist/flags/xx.png`. 

The impact:
* largely appearance; analytics continue to be recorded and viewable.

PR #275, opened 2023-03-13
Missing icons in Matomo install

jefferya:
Addresses https://github.com/Islandora-Devops/isle-buildkit/issues/274

* uses the Matomo advised download location that contains the Git Submodule contents whereas the Git Release lacks the Git Submodule contents. See this issue for details:  https://github.com/matomo-org/matomo/issues/16553


nigelgbanks:
Seems the container no longer starts after this change? See the attached logs.

[matomo.log](https://github.com/Islandora-Devops/isle-buildkit/files/10964890/matomo.log)

nigelgbanks (changes requested):
Seems there might be additional steps required. Not 100% sure yet, it might be that the additional composer installs are messing with it. 

When testing start without a volume to reproduce the failed state.

nigelgbanks (approved):
I did a diff locally between the two and noticed the missing composer files, adding those back in before installing extra-tools did the trick.

nigelgbanks:
Thanks, @jefferya!

PR #277, opened 2023-05-18
Matomo upgrade to latest 4.14.2

g7morris:
* This MR upgrade updates the following

* Matomo from `4.13.0` to latest `4.14.2`
* Extra Tools from  `4.0.1-beta4` to latest `4.1.0-beta5`
* Adds a new feature - free Geolocation database that endusers can use for more accurate results
  * located at `/var/www/matomo/misc` and depending on the month the resulting file will look like `/var/www/matomo/misc/dbip-city-lite-2023-05.mmdb`
  * To configure Matomo to use this file / geolocation, after setup one can run 
 
```bash
docker-compose exec matomo with-contenv bash -lc "mysql --user=root --password=`cat secrets/DB_ROOT_PASSWORD` --database=matomo --host=mariadb --port=3306 -A -e 'update matomo_option set option_value=\"geoip2php\" where option_name = \"usercountry.location_provider\";'" ``
```

PR #278, opened 2023-05-23
Matomo upgrade 4142

nigelgbanks:
Rebased to cleanup commit history from pull https://github.com/Islandora-Devops/isle-buildkit/pull/277 also made the source branch on this repository so github actions runs.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -29,6 +29,7 @@ RUN --mount=type=cache,id=base-apk-${TARGETARCH},sharing=locked,target=/var/cach
         curl \
         git \
         gnupg \
+        gzip \
         jq \
         mariadb-client \
         mysql-client \
--- a/base/rootfs/usr/local/bin/download.sh
+++ b/base/rootfs/usr/local/bin/download.sh
@@ -112,9 +112,6 @@ function unpack {
         args+=("--strip-components" "1")
     fi
     case "${file}" in
-    *.gz | *.gzip)
-        gunzip -fq "${file}" -c "${dest}"
-        ;;
     *.tar.xz | *.txz)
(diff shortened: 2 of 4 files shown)


nigelgbanks:
Sorry, travel got delayed, so I took another day to look at this. I reworked download.sh so it worked for the IP database, but in checking the Matomo docs and UI, it seems like the download of the database is an automatically managed feature. So, we don't need to download the database manually. Also, since it's tied to the month, we probably shouldn't bake this into the image directly, as it will rapidly go out of date?

https://matomo.org/faq/how-to/faq_163/

Note: The latest version of Matomo includes an auto-updater for these databases.

nigelgbanks:
@g7morris can you look into https://github.com/Islandora-Devops/isle-buildkit/pull/278#issuecomment-1560257747 and explain why it's required to manually download the database?

g7morris:
Hi @nigelgbanks 

Quick followup. Reviewed further and didn't like the hardcoded elements for the geoIP db which would require constant human intervention and updates. Refactored the code to download and update on every build. I think this would be nicer to do if we agree to bake into the image. 

Additionally this would negate any further need for a bind mount of that file. (ignore my comment on that particular bind mount above if we agreed)

I think we're ready to move forward now.

g7morris (changes requested):
Hi @nigelgbanks 

I've reviewed and updated as requested. 

Over to you again for review.

nigelgbanks:
Hi @g7morris I'll take a crack at this on Thursday. I should have sometime then. Or if I'm not too exhausted at the airport this morning.

g7morris:
I've now updated this to https://github.com/Islandora-Devops/isle-buildkit/tree/matomo-upgrade-4151

PR #281, opened 2023-06-13
issue-280 Fix for IP Address forwarding

### LINKED ISSUE #280: Matomo recording the same IP for all clients
When using Isle-buildkit Tag 2.0.0, regardless of the client IP used to access the Drupal site, Matomo is recording a 172.22.0.0 IP address as opposed to the client IP. Are others experiencing this?

Diagnostics:

$_SERVER['HTTP_X_FORWARDED_FOR'] is correct within the Matomo container. Verified by checking Matomo->Administration->Diagnotics->phpinfo -- $_SERVER['HTTP_X_FORWARDED_FOR'] = my_routable_public_ip_address. Traefik appears to  be properly forwarding the client IP in the header.

Checking to see if Matomo is setup correctly: 

```
$ ./console config:get General.proxy_client_headers
"HTTP_X_FORWARDED_FOR"
```

The above looks correct as per:
* <https://github.com/Islandora-Devops/isle-buildkit/tree/main/matomo>
* <https://matomo.org/faq/how-to-install/faq_98/>

Trying an experiment, let's set an array instead of a single value for `proxy_client_headers`:

* create array value via `"proxy_client_headers[]"` (note the `[]`)
```
$ ./console config:set --section="General" --key="proxy_client_headers[]" --value="${MATOMO_PROXY_CLIENT_HEADERS}"

$ ./console config:get General.proxy_client_headers
["HTTP_X_FORWARDED_FOR"]

```

Matomo is now recording the correct client IP (i.e., 172.22.0.0 private IP representing Docker)

A fix might be changing https://github.com/Islandora-Devops/isle-buildkit/blob/2173ab58eaff2d5575664fc2ea71765927743bc7/matomo/rootfs/etc/s6-overlay/scripts/install.sh#L56

To (untested) a version with `[]` as part of the key
```
    add_setting General "proxy_client_headers[]" "${MATOMO_PROXY_CLIENT_HEADERS}"
```

I think the reason the array works but not the single value assignment of `proxy_client_headers` is the the Matomo code is clobbering the value here (though I might be misunderstanding the usage of the `@` in this  context):
```
        $general = Config::getInstance()->General;
        $clientHeaders = @$general['proxy_client_headers'];
        if (!is_array($clientHeaders)) {
            $clientHeaders = array();
        }
```
https://github.com/matomo-org/matomo/blob/5505608c6504f80e5f4bcbc609370180489d420c/core/IP.php#L48-L52

nigelgbanks:
Resolves #280, pinging @jefferya.

### DIFF
--- a/matomo/rootfs/etc/s6-overlay/scripts/install.sh
+++ b/matomo/rootfs/etc/s6-overlay/scripts/install.sh
@@ -52,11 +52,11 @@ function install {
     s6-setuidgid nginx /var/www/matomo/console plugin:activate ExtraTools
 
     # Add additional configurations.
-    add_setting General assume_secure_protocol "${MATOMO_ASSUME_SECURE_PROTOCOL}"
-    add_setting General proxy_client_headers "${MATOMO_PROXY_CLIENT_HEADERS}"
-    add_setting General proxy_host_headers "${MATOMO_PROXY_HOST_HEADERS}"
-    add_setting General force_ssl "${MATOMO_FORCE_SSL}"
-    add_setting General proxy_uri_header "${MATOMO_PROXY_URI_HEADER}"
+    add_setting General "assume_secure_protocol" "${MATOMO_ASSUME_SECURE_PROTOCOL}"
+    add_setting General "proxy_client_headers[]" "${MATOMO_PROXY_CLIENT_HEADERS}"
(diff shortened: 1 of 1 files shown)


jefferya (approved):
Thank you! 👍

PR #285, opened 2023-08-10
Update alpine/drush launcher.

nigelgbanks:
Just some house keeping, needs testing.


rosiel:
Testing log:
1. download this PR in buildkit. Ran `docker buildx bake` to create all the images locally. (`make bake` failed, not sure why)
2. set up a new isle-site-template using the automatic method. Edit .env and change `ISLANDORA_TAG` from `main` to `local`. Also changed the site name to `test-drush`.
3. build the drupal container and start the containers using the commands in README.template.md.

Result:
drupal container failed and exited. 

Note: I think the testing was legit despite me not knowing what i'm doing, because the image used by the drupal container, named islandora.io/test-drush:local includes, in its layers,  drush version 0.10.2.

The problem I think is that drush launcher doesn't work with drush 12, which is what i'm trying to install with the starter site because it's the only one that's not facing down EOL in the next few months.

rosiel:
That said, I re-tested with 0.8.0 and it works great. drush runs in the ide. So this is a good bit of housekeeping though the launcher part is still a problem. Mind if i merge?

rosiel (approved):

PR #294, opened 2023-10-25
add mariadb-server-utils to mariadb image

joshdentremont:
after upgrading mariadb (when we update to a new buildkit TAG) we sometimes need to run mariadb-upgrade to fix our existing databases. That is not included in the image unless we add mariadb-server-utils.

With this change you should be able to run mariadb-upgrade on your container.

## To test
Try to run mariadb-upgrade in your mariadb container with and without this PR.

### DIFF
--- a/mariadb/Dockerfile
+++ b/mariadb/Dockerfile
@@ -10,6 +10,7 @@ RUN --mount=type=cache,id=mariadb-apk-${TARGETARCH},sharing=locked,target=/var/c
     apk add \
         mariadb \
         mysql-client \
+        mariadb-server-utils \
     && \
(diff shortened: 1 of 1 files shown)


joshdentremont:
Assuming this gets merged, can we also get a new tag for buildkit so that this is included? I have also added this to my documentation PR at https://github.com/Islandora/documentation/pull/2264

aOelschlager (approved):
This fixed the incorrect definition of table mysql.column_stats error I was getting after pulling the new images for an existing site.

aOelschlager:
A new tag will need to be made after this merge.

nigelgbanks (approved):

PR #305, opened 2023-11-17
Bumped fcrepo tool dependencies to latest versions.

nigelgbanks:


nigelgbanks:
@joshdentremont I've updated everything to the latest if you want to to test.

joshdentremont:
We just discussed this at the tech call. I don't have the capacity to test this at the moment, but wanted to add some notes for someone to test if they have the chance.

This PR is updating:
- Fedora
- Syn
- Import/Export
- Upgrade Utils

For the upgrade utils we should probably test an import from Fedora 5 to Fedora 6 because of the new OCFL updates that are included in this. In order to do this I think we need to:
1. Spin up a new site with Fedora 5
2. Add some data to Fedora
3. Export the Fedora data
4. Update Fedora to 6 or start another new site with Fedora 6
5. Import the data that was exported from Fedora 5
6. Run the upgrade utils and make sure the import was successful

The import / export process is described in the Isle-dc makefile here: https://github.com/Islandora-Devops/isle-dc/blob/development/Makefile#L418-L448 

That should cover the testing of import/export, and of upgrade utils. 

I'm not sure if further testing is needed for Fedora and Syn, or if the above covers that as well.

nigelgbanks:
This no longer contains code changes, as I forgot about it when upgrading all other components in https://github.com/Islandora-Devops/isle-buildkit/pull/330. This only now updates the documentation on how to update Fcrepo dependencies, so I'm just going to push this in since it's been open for 5 months.

PR #306, opened 2023-11-17
changed matomo volume to only persist config directory

joshdentremont:
Having the Matomo volume persist the entire /var/www/matomo directory causes problems when upgrading Matomo. This change should allow us to persist the config directory, which I believe is the only thing that needs to persist. 

With this, when new Matomo images are created with new versions of Matomo, we should only need to run database updates. 

This should prevent this issue: https://github.com/Islandora-Devops/isle-dc/issues/361

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -276,7 +276,7 @@ services:
         environment:
             MATOMO_DEFAULT_HOST: "https://islandora.dev"
         volumes:
-            - matomo-data:/var/www/matomo:rw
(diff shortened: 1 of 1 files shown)


nigelgbanks:
Not a huge concern for isle-build kit since it's only for testing. Though I think for isle-dc, and others it might be a problem as users can install plugins via the web-ui

nigelgbanks (approved):

Issue #307, opened 2023-11-20
It would be really nice to be able to set Heap sizes for Java in Cantaloupe


noahwsmith:
I think it would go here:

https://github.com/Islandora-Devops/isle-buildkit/blob/main/cantaloupe/rootfs/etc/s6-overlay/s6-rc.d/cantaloupe/run

Having `CANTALOUPE_HEAP_MIN` and `CANTALOUPE_HEAP_MAX` variables that passed into `Xms` and `Xmx` parameters in that command would do the trick.

g7morris:
Hello everyone, 

Today I was able to:

* create a new MR based on the changes suggested in this ticket - https://github.com/Islandora-Devops/isle-buildkit/pull/310/files

* run `make bake cantaloupe` on isle-buildkit without failure resulting in a new Docker image.

* test an isle-dc based project using this new image (local) by
  * using a different tag for cantaloupe - `    image: ${REPOSITORY}/cantaloupe:local`
  * adding the following to the cantaloupe `environment` section of the docker-compose.yml

```bash
      CANTALOUPE_HEAP_MIN: "3G"
      CANTALOUPE_HEAP_MAX: "5G"
```

   * increase and decrease the heap values dynamically. I increased to 4 GB min and 6 GB max.
     * Reviewed change in values by `docker-compose exec cantaloupe bash -lc "printenv | grep HEAP"`

Test 1

```bash
CANTALOUPE_HEAP_MAX=5G
CANTALOUPE_HEAPCACHE_PERSIST=false
CANTALOUPE_HEAPCACHE_PERSIST_FILESYSTEM_PATHNAME=/data/heap.cache
CANTALOUPE_HEAPCACHE_TARGET_SIZE=2G
CANTALOUPE_HEAP_MIN=3G
```

Test 2
```bash
CANTALOUPE_HEAP_MAX=6G
CANTALOUPE_HEAPCACHE_PERSIST=false
CANTALOUPE_HEAPCACHE_PERSIST_FILESYSTEM_PATHNAME=/data/heap.cache
CANTALOUPE_HEAPCACHE_TARGET_SIZE=2G
CANTALOUPE_HEAP_MIN=4G
```

@noahwsmith and @nigelgbanks Anything else I should do to test or consider?

g7morris:
Possibly add these values to the [isle-dc sample.env](https://github.com/Islandora-Devops/isle-dc/blob/development/sample.env#L200) once this MR is merged?

[...]

PR #310, opened 2023-11-28
Issue 307 - Adding xms and xmx for Cantaloupe heap

g7morris:
Per https://github.com/Islandora-Devops/isle-buildkit/issues/307


noahwsmith (approved):
I've tested this in our base project, the variables do what we want and things work pretty well. I snuck in adding a missing library `libjpeg-turbo`.

noahwsmith:
You've set default values so it technically doesn't require a change, but folks should know that the defaults have been set and how to change them, for sure.

aOelschlager (approved):
Tested this in isle-dc. I pulled in the 307-set-heap-sizes-for-java-in-cantaloupe images and ran `ps aux | grep java` in the cantaloupe container which displayed the default values for xms and xmx. I then used the xms and xmx variables in my docker compose file but set the values a different amount than the default values. I restarted the containers and I ran the same command as before and it displayed the values from my docker compose file.

PR #313, opened 2023-12-01
Log4j logging files in Solr are too old. Replacing with latest to avoid security concerns

g7morris:
WIP for upgrading LOG4J logging files that are 2.17.x vs latest 2.22.0. This is a concern for security and CVEs. Will test shortly.


nigelgbanks:
@joecorall raised a point that we could upgrade to Solr 9.x, maybe that's the route forward? https://islandora.slack.com/archives/C03472QEUM8/p1701358102239509

g7morris:
Okay I'll close this PR then as upgrading to Solr 9.x sounds like the community solution. @nigelgbanks or @joecorall Are either of you opening up a new issue to this effect and won't the current Drupal schema be impacted by this?

PR #325, opened 2024-02-14
Remove obsolete Crayfish parameter syn_config.

alxp:
Related GitHub issue:

https://github.com/Islandora/Crayfish/issues/181

From my understanding of Crayfish's UPGRADE.md file, this config key is no longer used and its presence is causing errors with Hypercube.

https://github.com/Islandora/Crayfish/blob/4.x/Hypercube/UPGRADE.md

It looks like the other config changes needed for the upgrade are already in place.


jefferya:
Tested locally. Fixes problem.
My test:
1. Reproduce error: use Tag 3.1.2 and import a PDF via Islandora Workbench;
    * results: hypercube "syn_config" error and no OCR media
2. Update to image produces by this PR and import PDF via Islandora Workbench; 
    * results: no errors in the hypercube log; OCR media generated

PR #332, opened 2024-05-09
Log4j cleanup on Blazegraph and FITS

noahwsmith:
Clean out ancient log4j and replace with patched version.


noahwsmith:
Updated to include cleanup on FITS and Solr as well as Blazegraph. @misilot has offered to test.

noahwsmith:
The built test images are available as:
`borndigital/blazegraph:blazegraph-log4j`, `borndigital/solr:blazegraph-log4j` and `borndigital/fits:blazegraph-log4j`

noahwsmith:
Backing out of Solr - not necessary https://github.com/apache/solr-docker/blob/main/9.5/Dockerfile
And covered here by the update to Solr 9.x: https://github.com/Islandora-Devops/isle-buildkit/pull/313

solr/Dockerfile
```diff
@@ -7,6 +7,12 @@ ARG SOLR_FILE=solr-${SOLR_VERSION}.tgz
 ARG SOLR_URL=https://archive.apache.org/dist/solr/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz
 ARG SOLR_FILE_SHA256=d8538502019af1945e0b124a4613b46ca43aedcf3f20e9912c482c080407ea21
 
+ARG LOG4J_VERSION="2.22.0"
+ARG LOG4J_FILE="apache-log4j-${LOG4J_VERSION}-bin.zip"
```
misilot, line 11: @noahwsmith do you want to keep these in here as well for Solr, or should they come out as well.

misilot:
@noahwsmith it looks like it might be still be part of a couple of layers :(

Path              : /var/lib/docker/overlay2/bfee10e74d13452ab0bb6c93123be38cea006ec7d705911b6db71d53c4e339c1/diff/opt/tomcat/webapps/bigdata/WEB-INF/lib/log4j-1.2.17.jar
  Installed version : 1.2.17

  Path              : /var/lib/docker/overlay2/e9f42b9942914e758fa4c1e32cd87794de5c3274a1e70a56a5f901d742747176/diff/opt/tomcat/webapps/bigdata/WEB-INF/lib/log4j-1.2.17.jar
  Installed version : 1.2.17

blazegraph/Dockerfile
```diff
@@ -16,6 +21,21 @@ RUN --mount=type=cache,id=blazegraph-downloads-${TARGETARCH},sharing=locked,targ
...
     cleanup.sh
 
+# Remove old files & then install latest log4j-* files
+RUN --mount=type=cache,id=log4j-downloads-${TARGETARCH},sharing=locked,target=/opt/downloads \
+    download.sh \
+        --url "${LOG4J_URL}" \
+        --sha256 "${LOG4J_FILE_SHA256}" \
+    && \
+    ## Remove the outmoded log4j-* files that come with blazegraph
+    rm -f "/opt/tomcat/webapps/bigdata/WEB-INF/lib/log4j-1.2.17.jar" && \
```
noahwsmith, line 31: This line specifically removes `/opt/tomcat/webapps/bigdata/WEB-INF/lib/log4j-1.2.17.jar`

noahwsmith: @misilot The way this PR works is simply that we scrub that out after it is initially brought in by the Blazegraph install. So it's not surprising to me that the file is in an earlier layer but it is not in the final image.

solr/Dockerfile
noahwsmith: Oops I'll fix that too - thanks

blazegraph/Dockerfile
misilot: I think Line 11/16 is adding it. The two RUN commands I think need to be combined, so there isn't a layer where the log4j*.jar exists.

noahwsmith: Good idea - I'll try that...

noahwsmith: Ok, building now. I brought the RM back up to the first RUN...

misilot:
Thanks @noahwsmith this PR seems to work great, and we are no longer getting the log4j hits on our scans.

noahwsmith:
@joecorall How do you feel about this? Any objections to merging?

joecorall (approved):

Issue #333, opened 2024-05-13
Being able to set CORS with a .env variable would be nice


noahwsmith:
To support inter-institutional collaboration, sometimes setting non-default values for CORS is necessary. Currently to do this requires bindmounting a copy of https://github.com/Islandora-Devops/isle-buildkit/blob/main/drupal/rootfs/etc/confd/templates/drupal.fpm.conf.tmpl and inserting into that template. 

I believe that this is enough of a "standard use case" that we should support it with a `.env` variable. I will issue a PR momentarily to address this.

noahwsmith:
What I've done is the VERY SIMPLE version that lets you set a single domain, or `*`. We could also look at a more complex situation where you pass in a regex that sets a list of allowed domains since you cannot just give it a list in string format. For example: https://github.com/kubernetes/ingress-nginx/issues/5496

noahwsmith:
See comment in https://github.com/Islandora-Devops/isle-buildkit/pull/334 for alternate solution inside of Drupal.

PR #334, opened 2024-05-13
add DRUPAL_CORS variable

noahwsmith:
Addresses https://github.com/Islandora-Devops/isle-buildkit/issues/333

The goal here is to add a .env variable that will allow setting a CORS value for Drupal without bindmounting the drupal.fpm.conf.tmpl file.

### DIFF
--- a/drupal/README.md
+++ b/drupal/README.md
@@ -27,6 +27,7 @@ additional settings, volumes, ports, etc.
 | :----------------------- | :------ | :--------------------------------------------------------------------------------- |
 | DRUPAL_ENABLE_HTTPS      | true    | Inform PHP that `https` should be used.                                            |
 | DRUPAL_REVERSE_PROXY_IPS |         | Use the IP address for the host 'traefik' if found otherwise default to `0.0.0.0`. |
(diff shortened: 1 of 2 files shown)


joecorall:
@noahwsmith - is this needed for static assets? Or can https://www.drupal.org/node/2715637 be used?

noahwsmith:
Hi Joe- I tested that and it turned out it worked well. I'll close out this Issue and PR. 

We had previously had issues with multiple domains, but the current cors service syntax now allows what we need. For posterity, here is a working config that folks can adopt if they find this issue and have a similar need:

```
parameters:
  cors.config:
    enabled: true
    # Configure requests allowed from specific origins by commenting this out. All domains by uncommenting.
    #allowedOrigins: ['*']
    # Configure requests allowed from multiple origins matching against regex patterns.
    # You shouldn't allow "*" above if you're using Patterns also. Example patterns:
    allowedOriginsPatterns: ['/(.*\.traefik\.me|.*\.sitedomain\.edu|mirador.*\.netlify\.app)/']
```

PR #335, opened 2024-05-31
Add additional environment variables to better support reverse proxies

joecorall:
If your ISLE deployment is behind a reverse proxy, several IPs will be hit before passing the request along to Drupal. This PR simply allows an admin to specify those proxies so the log entries in drupal's nginx container will have the original client IP


ajstanley (approved):
Looks good.

PR #336, opened 2024-06-03
[major] Remove matomo

joecorall:
[In May 2024 it was announced Islandora would no longer ship with Matomo](https://groups.google.com/g/islandora/c/MtHOxjew-tA/m/b_UWDwdJBwAJ).

> However, they [matomo] will not be maintained by the Foundation after 6 months and will be removed after 1 year.

We promised to maintain the matomo image for six months, which we're past at this point. We can keep the old images around on the 3.x tags as promised until June 2025 (the 3.x tags will probably last past then which is 100% OK)

This will bump buildkit to 4.x


nigelgbanks:
Relevant discussion https://islandora.slack.com/archives/CM5PPAV28/p1715949100627349

nigelgbanks:
I won't have my laptop till Monday, so I'm a bit limited. But I've noticed there are other references to Matomo, to be removed. 

Though I would suggest that we completely remove Matomo and do a major release instead. If organizations want to upgrade they can continue to use the previous release of the Matomo image.

PR #337, opened 2024-06-03
Add Drupal 10+ CSS/JS Aggregation support

joecorall:
With CSS/JS aggregation enabled in Drupal 10, nginx needs to first check if the CSS/JS exists on disk. If not, needs to forward the request to Drupal/php

## How to test

Using `ISLANDORA_TAG=3.2` see that when CSS/JS aggregation is enabled in your Drupal stack, your theme's page displays are broken

Using this PR's tag `ISLANDORA_TAG=d10-css-js-aggregations` see that CSS/JS aggregation works as expected

## Related conversations

https://islandora.slack.com/archives/C019U12D44Q/p1717446855079619?thread_ts=1699377978.424099&cid=C019U12D44Q

### DIFF
--- a/drupal/rootfs/etc/nginx/shared/drupal.defaults.conf
+++ b/drupal/rootfs/etc/nginx/shared/drupal.defaults.conf
@@ -85,6 +85,11 @@ location ~ ^(/[a-z\-]+)?/system/files/ { # For Drupal >= 7
     try_files $uri /index.php?$query_string;
 }
 
+# handle CSS/JS aggregation through Drupal
+location ~ ^/sites/default/files/(css|js)/ {
+    try_files $uri /index.php?$query_string;
+}
+
 location ~* \.(js|css|png|jpg|jpeg|gif|ico|svg)$ {
     try_files $uri @rewrite;
     expires max;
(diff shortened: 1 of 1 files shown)


drupal/rootfs/etc/nginx/shared/drupal.defaults.conf
```diff
@@ -85,6 +85,11 @@ location ~ ^(/[a-z\-]+)?/system/files/ { # For Drupal >= 7
     try_files $uri /index.php?$query_string;
 }
 
+# handle CSS/JS aggregation through Drupal
+location ~ ^/sites/default/files/(css|js)/ {
```
nigelgbanks, line 89: Can you replace "default" with ".*" so that aggregation can work with multi-sites.

nigelgbanks: Or it could be merged with line 78, which was the previous support for styles generation.

nigelgbanks (changes requested):
Looks good other than the one issue I highlighted.

drupal/rootfs/etc/nginx/shared/drupal.defaults.conf
joecorall: Good call 👍 https://github.com/Islandora-Devops/isle-buildkit/pull/337/commits/a6916e6752ad152fd7031a5f23da0d7e7ca8df5a

nigelgbanks (approved):

PR #338, opened 2024-06-04
Solr: Add OCR Highlight plugin.

alxp:
The build will fail if there is a problem downloading the file, otherwise you can just look for the presence of the downloaded .jar file in the above location.## What does this pull request do?

Adds the [OCR Highlighting](https://github.com/dbmdz/solr-ocrhighlighting) plugin to the Solr image.

The jar is placed in a directory where Solr won't find it unless it is explicitly configured to do so, which Drupal sites that are set up to support Solr and hOCR will do. But otherwise the presence of the .jar in the 

/opt/solr/server/solr/contrib/ocrhighlighting/lib/solr-ocrhighlighting-0.8.6.jar

folder will have no effect.

### DIFF
--- a/solr/Dockerfile
+++ b/solr/Dockerfile
@@ -8,7 +8,7 @@ ARG SOLR_URL=https://archive.apache.org/dist/solr/solr/${SOLR_VERSION}/solr-${SO
 ARG SOLR_FILE_SHA256=d8538502019af1945e0b124a4613b46ca43aedcf3f20e9912c482c080407ea21
 ARG OCRHIGHLIGHT_VERSION=0.8.6
 ARG OCRHIGHLIGHT_FILE=solr-ocrhighlighting-${OCRHIGHLIGHT_VERSION}.jar
-ARG OCRHIGHLIGHT_URL=https://github.com/dbmdz/solr-ocrhighlighting/releases/download/0.8.6/solr-ocrhighlighting-${OCRHIGHLIGHT_VERSION}.jar
+ARG OCRHIGHLIGHT_URL=https://github.com/dbmdz/solr-ocrhighlighting/releases/download/${OCRHIGHLIGHT_VERSION}/solr-ocrhighlighting-${OCRHIGHLIGHT_VERSION}.jar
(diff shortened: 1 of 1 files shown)


solr/Dockerfile
```diff
@@ -6,6 +6,11 @@ ARG SOLR_VERSION=9.5.0
 ARG SOLR_FILE=solr-${SOLR_VERSION}.tgz
 ARG SOLR_URL=https://archive.apache.org/dist/solr/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz
 ARG SOLR_FILE_SHA256=d8538502019af1945e0b124a4613b46ca43aedcf3f20e9912c482c080407ea21
+ARG OCRHIGHLIGHT_VERSION=0.8.6
+ARG OCRHIGHLIGHT_FILE=solr-ocrhighlighting-${OCRHIGHLIGHT_VERSION}.jar
+ARG OCRHIGHLIGHT_URL=https://github.com/dbmdz/solr-ocrhighlighting/releases/download/0.8.6/solr-ocrhighlighting-${OCRHIGHLIGHT_VERSION}.jar
```
nigelgbanks, line 11: There is a literal version number here that needs to be replaced by the variable above.

solr/Dockerfile
```diff
@@ -24,6 +29,15 @@ RUN --mount=type=cache,id=solr-downloads-${TARGETARCH},sharing=locked,target=/op
     && \
     cleanup.sh
 
+RUN --mount=type=cache,id=solr-ocrhighlight-downloads,sharing=locked,target=/opt/downloads \
+    mkdir -p ${OCRHIGHLIGHT_DEST} && \
```
nigelgbanks, line 33: Not sure if this is actually needed. I think the download script will do so. If that's the case we can get rid of the variable for the directory since it will only be used on one spot.

nigelgbanks (changes requested):
Aside from the two issues I highlighted, this looks ready to go in.

nigelgbanks (approved):

Issue #260, opened 2022-11-01
Deprecate Drupal utility scripts


nigelgbanks:
We should be pushing for images that are by and large configured at build time or in the worst case by `confd` templates, and should not rely on running commands via `docker exec`. As it is unreliable and less predictable and requires a release of `isle-buildkit` for downstream consumers. 

To that end:

- An audit should be done of all the scripts
- A deprecate notice should be raised (not sure how to do this as they are bash scripts, but it should go into the release notes at least).
- Update the documentation to describe a better method for applying configuration from environment variables (i.e. how sandbox does it).
- On a subsequent release, the scripts can then be removed.

For example:
https://github.com/Islandora-Devops/isle-buildkit/blob/d5e177dc50acf964925f59684b6988f25a2620e2/drupal/rootfs/etc/islandora/utilities.sh

Has many assumptions about modules configuration etc, and is used to mutate `settings.php` at run time.

Those scripts were born out of porting what Ansible did into Docker, a good first start, but it's time to move on from them.

joecorall:
> Update the documentation to describe a better method for applying configuration from environment variables (i.e. how sandbox does it).

This might be able to help here: https://github.com/islandora-devops/islectl

We can carve out common tasks as commands on that utility. If we identify what those are i can build them out. Or anyone comfortable with go PRs are welcome.

PR #349, opened 2024-09-09
hard code site root

joshdentremont:
This should fix the issue introduced by drush 13 that stopped us from running certain drush commands before the site was installed. See https://github.com/Islandora-Devops/isle-site-template/pull/58 for more details.

AFAIK the drupal root is always the same when running buildkit in either Isle-dc or site template so this should be fine. 

To test: 
- build the drupal image in this PR with `docker buildx bake drupal`
- create a new site with Site Template as normal
- change the Dockerfile in Isle Site Template to say `FROM islandora/drupal:local` so that it builds from this image
- build a new Drupal image in the site template
- spin up site using `docker compose --profile dev up -d` as normal

### DIFF
--- a/drupal/rootfs/etc/islandora/utilities.sh
+++ b/drupal/rootfs/etc/islandora/utilities.sh
@@ -388,7 +388,7 @@ function update_settings_php {
         fedora_url previous_owner_group
     site="${1}"
     shift
-    drupal_root=$(drush drupal:directory)
+    drupal_root=/var/www/drupal/web
     site_url=$(drupal_site_env "${site}" "SITE_URL")
     driver=$(drupal_site_env "${site}" "DB_DRIVER")
     host=$(drupal_site_env "${site}" "DB_HOST")


aOelschlager:
I would add that there are other places where [drupal:directory is referenced in the utilities.sh](https://github.com/search?q=repo%3AIslandora-Devops%2Fisle-buildkit%20drupal%3Adirectory&type=code) script that should be changed as well, in particular the update_settings_php that broke the isle-dc installation (has had fix with this [pr](https://github.com/Islandora-Devops/isle-dc/pull/403))

Could the calls made to the drupal:directory could be replaced with an environment variable instead?

joshdentremont:
That drush command works once the site is installed, so everything else seems fine on my site template test so far. 

Not sure about isle-dc, though. If it tries to run update_settings_php before doing a site install then it would likely fail, but I would think that should only run on installed sites.

nigelgbanks (approved):

nigelgbanks:
> I would add that there are other places where [drupal:directory is referenced in the utilities.sh](https://github.com/search?q=repo%3AIslandora-Devops%2Fisle-buildkit%20drupal%3Adirectory&type=code) script that should be changed as well, in particular the update_settings_php that broke the isle-dc installation (has had fix with this [pr](https://github.com/Islandora-Devops/isle-dc/pull/403))
> 
> Could the calls made to the drupal:directory be replaced with an environment variable instead?

There is an open issue to deprecate that whole file. 

https://github.com/Islandora-Devops/isle-buildkit/issues/260

nigelgbanks:
The others need to be changed as well... https://github.com/Islandora-Devops/isle-buildkit/pull/350 has the required changes along with a few other bugs I found while looking into this.

PR #350, opened 2024-09-12
Update starter site & matomo extra tools & fix permission Bugs

nigelgbanks:
A number of permission issues have gone unnoticed,
I became aware of them when I did a fresh build
without cache on a newly installed computer.

When building Docker images, Docker copies
files/folders with the same permissions as they
have on the host machine. Git does not track the
permissions of files except for the executable bit
on files. So depending on who builds the images
files/folders can get different permissions.

This changes it such that the Makefile will ensure
the appropriate permissions are set when building.

This is required as we run services as different
users other than root in the container. Those
users must be able to access and execute certain
files.


aOelschlager (approved):
I tested this with a new and existing site on isle-dc and it works. I was not able to test Matomo because the current starter site no longer uses it. I could pull from an older starter site commit I believe, if that is needed to be tested.

nigelgbanks:
@aOelschlager thanks! That should be enough testing 😄

PR #358, opened 2024-12-28
integrate with renovate

joecorall:
After putting together the PR for https://github.com/Islandora-Devops/isle-buildkit/pull/357 yesterday I wondered if [renovate](https://github.com/renovatebot/renovate) could be used to help find/detect/issue PRs for dependencies in our Dockerfiles. renovate has good support for github tags/releases, which most of our dependencies live at. We can write custom managers for other dependencies, but first thought I'd send this over to see if this is how we want to proceed (i.e. do we want to use renovate for this, or maybe some other tool).

Basically, if this PR is merged we can follow [the instructions in the README](https://github.com/Islandora-Devops/isle-buildkit/blob/4aab1a2047dd7c161aec655b345f920092716d01/README.md#renovate) to update a whole slew of dependencies. If we're in agreement renovate is the preferred approach for this, we can add other dependencies to be checked, and look into automating the PR process. Right now a PR would need created by manually running the renovate CLI on a developer's workstation with the proper dependencies installed and [the proper GitHub token](https://docs.renovatebot.com/modules/platform/github/#running-using-a-fine-grained-token). Automating could just be a github action we add the properly scoped token for in a secret that runs on a schedule. We probably should use a robot github account for that.

The PR that will get issued in this repo will look like this: https://github.com/joecorall/isle-buildkit/pull/12


nigelgbanks (approved):
This looks fine to me as far as I can tell. I tried running it, following the instructions in the README. It produced this for some reason: https://github.com/Islandora-Devops/isle-buildkit/pull/359, I suspect because this is not merged yet.

joecorall:
> I suspect because this is not merged yet.

Yeah, the renovate JSON config needs to be on the main branch.

PR #366, opened 2025-01-04
Automatically run renovate on a schedule

joecorall:
I created a GitHub account: https://github.com/isle-buildkit-renovate

Then gave that account access to only this repo. Then generated a PAT for that account and added it as a GitHub Actions secret in this repo. That PAT would need renewed every year.

If this PR is accepted/merged, renovate will bump any dependencies it finds either in a manual run of the proposed GitHub workflow, or automatically every Sunday.

Creating the `isle-buildkit-renovate` GitHub bot account was five minutes of work, so easy to undo if we want to use something else. If we'd rather use a different account for this, let's just discuss in this PR or in Islandora Slack. The email for the bot account -- isle-buildkit-renovate@libops.io -- is just a Google Group anyone can be added to for "forgot my password" or any other purposes. We can also just update the GitHub account's email to something else if desired.

### DIFF
--- a/.github/workflows/renovate.yml
+++ b/.github/workflows/renovate.yml
@@ -1,0 +1,26 @@
+name: run renovate
+
+on:
+  workflow_dispatch:
+  # Monday mornings
+  schedule:
+    - cron: '15 1 * * 1'
+
+env:
+  LOG_LEVEL: debug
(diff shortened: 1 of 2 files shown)


joecorall:
huh I just noticed https://github.com/actions/create-github-app-token 🤔

PR #376, opened 2025-02-07
Add code server healthcheck

joecorall:
Since the image base is drupal, the code server container had a healthcheck from nginx (and wasfailing). Instead, we should hit the `/healthz` endpoint in the codeserver.


joecorall:
I am going to go ahead and merge this since the `main` tag will cause dev environments to break with code server having a bad healthcheck (that causes traefik to drop the codeserver)

Issue #377, opened 2025-02-09
Label images with opencontainers annotations


xurizaemon:
Automated tools such as Renovate consult image labels to identify image provenance. By adding labels we may be able to better support use of those tools, eg Renovate would be better equipped to provide changelogs when proposing updates to Islandora images.

- [Renovate Docker datasource docs](https://docs.renovatebot.com/modules/datasource/docker/)
- [OpenContainers annotations spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md)
- Follow-up to #373

There are plenty of labels we _could_ add - I don't want to spam the Dockerfiles, so I'll kick off a PR with only a couple.

| Label key | Description ([source](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys)) |
|--|--|
| org.opencontainers.image.created | date and time on which the image was built, conforming to RFC 3339. |
| org.opencontainers.image.authors | contact details of the people or organization responsible for the image (freeform string) |
| org.opencontainers.image.url | URL to find more information on the image (string) |
| org.opencontainers.image.documentation | URL to get documentation on the image (string) |
| org.opencontainers.image.source | URL to get source code for building the image (string) |
| org.opencontainers.image.version | version of the packaged software |
| org.opencontainers.image.revision | Source control revision identifier for the packaged software. |
| org.opencontainers.image.vendor | Name of the distributing entity, organization or individual. |
| org.opencontainers.image.licenses | License(s) under which contained software is distributed as an SPDX License Expression. |
| org.opencontainers.image.ref.name | Name of the reference for a target (string). |
| org.opencontainers.image.title | Human-readable title of the image (string) |
| org.opencontainers.image.description | Human-readable description of the software packaged in the image (string) |
| org.opencontainers.image.base.digest | Digest of the image this image is based on (string) |
| org.opencontainers.image.base.name | Image reference of the image this image is based on (string) |

We may also want to set the image `Created` property (as well as `org.opencontainers.image.created` Label?), but note discussion of caching in [Slack thread](https://islandora.slack.com/archives/CM6F4C4VA/p1739151608450929).

xurizaemon:
Let me know if any other labels would be appreciated, happy to add them to the PR!

joecorall:
I commented in the PR https://github.com/Islandora-Devops/isle-buildkit/pull/378#issuecomment-2648088069

xurizaemon:
[...]

PR #378, opened 2025-02-09
isle-buildkit#377: Add opencontainers annotations for image source, URL

xurizaemon:
Add opencontainers annotations to images

Refs #377

Generated with:
```bash
for DOCKERFILE in */Dockerfile ; do DIR=$( dirname $DOCKERFILE ) ; sed -i '3 a LABEL org.opencontainers.image.url=https://github.com/Islandora-DevOps/isle-buildkit/\n' $DOCKERFILE ; sed -i "3 a LABEL org.opencontainers.image.source=https://github.com/Islandora-DevOps/isle-buildkit/tree/main/$DIR" $DOCKERFILE ; done
```

### DIFF
--- a/activemq/Dockerfile
+++ b/activemq/Dockerfile
@@ -1,6 +1,9 @@
 # syntax=docker/dockerfile:1.5.1
 FROM java
 
+LABEL org.opencontainers.image.source=https://github.com/Islandora-DevOps/isle-buildkit/tree/main/activemq
+LABEL org.opencontainers.image.url=https://github.com/Islandora-DevOps/isle-buildkit/
+
 ARG TARGETARCH
 # renovate: datasource=github-tags depName=apache-activemq packageName=apache/activemq
 ARG ACTIVEMQ_VERSION=5.18.6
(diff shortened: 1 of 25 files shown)


xurizaemon:
Looks like this could be done via https://docs.docker.com/build/bake/reference/#targetannotations if that's preferred?

joecorall:
I think these labels are useful to add and this is a good start. Though I do not think they will help with a changelog in systems like renovate - for that we should work on making better release notes in our GitHub releases and folks should using renovate should use their github release data source to get updates.

We probably need to add some build args to populate some of the values in the proposed labels, as the source of the image isn't always the `main` branch (e.g. when building images off a PR).

nigelgbanks:
I'm afraid that using build arguments for annotations like this will invalidate caching and force rebuilds for images that are otherwise identical. I'd prefer to rely on https://docs.docker.com/build/metadata/attestations/slsa-provenance/ as they are not baked into the image and denote many aspects of how the image was built. This is not currently implemented in our build process, but I've tested it locally, and I am integrating it into other work that I'm doing.

xurizaemon:
OK, I'll close this if you have existing WIP on it. Would be happy to look at a Bake approach of that's helpful, please say so if it is!

PR #379, opened 2025-02-09
Remove no-op multistage builds from Dockerfiles

xurizaemon:
NB: I may be misunderstanding the intended outcome so please verify my proposed change carefully 😄 

AFAICT, a `FROM` statement without a following ` as <name>`, as seen here, will have no effect on the built image.  (I expect the image gets _pulled_, but that seems like a side-effect here, not an output?)

So - I think these extra FROM are redundant and can be discarded. My understanding is that if they were followed by a name (`FROM nodejs as node`) and had separate build instructions, that would be a multi-stage build, but that as-is these are a multi-stage build with a single output, and that the state is cleared when the second FROM command is reached.

- https://docs.docker.com/reference/dockerfile/#from

### DIFF
--- a/code-server/Dockerfile
+++ b/code-server/Dockerfile
@@ -1,5 +1,4 @@
 # syntax=docker/dockerfile:1.5.1
-FROM nodejs
 FROM drupal
 
 ARG TARGETARCH
--- a/houdini/Dockerfile
+++ b/houdini/Dockerfile
@@ -1,5 +1,4 @@
 # syntax=docker/dockerfile:1.5.1
-FROM imagemagick
 FROM crayfish
 
 ARG TARGETARCH
--- a/hypercube/Dockerfile
+++ b/hypercube/Dockerfile
@@ -1,5 +1,4 @@
 # syntax=docker/dockerfile:1.5.1
-FROM leptonica
 FROM crayfish AS hypercube
 
 ARG TARGETARCH


joecorall (changes requested):
These aren't as they appear, as they're getting aliased in docker-bake.hcl e.g. https://github.com/Islandora-Devops/isle-buildkit/blob/main/docker-bake.hcl#L315-L322

They get built in separate repos e.g. https://github.com/Islandora-Devops/isle-nodejs

Semi-related: I have plans to get these auto-updated using renovate and have the `FROM` in dockerfiles be better documented avoiding confusion such as this PR, but haven't had time to put up that PR yet. I started a while ago over at https://github.com/joecorall/isle-buildkit but it's nowhere close to ready. I should have time to push that forward this week

joecorall:
Also, the images do get used in the build e.g https://github.com/Islandora-Devops/isle-buildkit/blob/50fa27d500bd986c321e8ff3cf56dc6bbf50822f/code-server/Dockerfile#L38

nigelgbanks:
> Semi-related: I have plans to get these auto-updated using renovate and have the `FROM` in dockerfiles be better documented avoiding confusion such as this PR, but haven't had time to put up that PR yet. I started a while ago over at https://github.com/joecorall/isle-buildkit but it's nowhere close to ready. I should have time to push that forward this week

@joecorall Before you sink time into that, those repositories are likely to be merged into the isle-buildkit. I've solved a number of the caching issues, which was the reason they were split into their own repositories in the first place. I'm still working away it maybe another week or two before I'm ready with a pull request, though.

xurizaemon:
Thanks so much for documenting and closing this, and apologies for any bother! Much appreciate the insight.

PR #380, opened 2025-02-11
Pin apk packages

joecorall:
We can use renovate's [repology datasource ](https://docs.renovatebot.com/modules/datasource/repology/) to update pinned apk packages.

Renovate basically looks for updates at https://pkgs.alpinelinux.org/packages?name=&branch=v3.20 and will bump the respective version in our Dockerfiles along with our other dependency updates. This will give us a little more visibility into what all changes version to version when we cut new releases.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -21,7 +21,7 @@ LABEL License="MIT License"
 # Start s6
 ENTRYPOINT [ "/init" ]
 
-ENV \
+ARG \
   # renovate: datasource=repology depName=alpine_3_20/bash
   BASH_VERSION=5.2.26-r0 \
   # renovate: datasource=repology depName=alpine_3_20/curl
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -21,28 +21,64 @@ LABEL License="MIT License"
 # Start s6
 ENTRYPOINT [ "/init" ]
 
+ENV \
+  # renovate: datasource=repology depName=alpine_3_20/bash
+  BASH_VERSION=5.2.26-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/curl
+  CURL_VERSION=8.11.1-r1 \
+  # renovate: datasource=repology depName=alpine_3_20/git
+  GIT_VERSION=2.45.3-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/gnupg
+  GNUPG_VERSION=2.4.5-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/gzip
+  GZIP_VERSION=1.13-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/jq
+  JQ_VERSION=1.7.1-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/mariadb-client
+  MARIADB_CLIENT_VERSION=10.11.10-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/mysql-client
+  MYSQL_CLIENT_VERSION=10.11.10-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/netcat-openbsd	
+  NETCAT_OPENBSD_VERSION=1.226-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/openssl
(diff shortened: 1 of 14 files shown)


drupal/Dockerfile
```diff
@@ -9,9 +9,6 @@ WORKDIR /var/www/drupal
 
 # Platform specific does require arch specific identifier.
 RUN --mount=type=cache,id=drupal-apk-${TARGETARCH},sharing=locked,target=/var/cache/apk \
-    apk add \
-      patch \
```
joecorall, line 13: I removed this since it's in the base image

mariadb/Dockerfile
```diff
@@ -5,12 +5,15 @@ ARG TARGETARCH
...
 
+ENV \
+  # renovate: datasource=repology depName=alpine_3_20/mariadb
+  MARIADB_VERSION=10.11.10-r0
+
 # Platform specific does require arch specific identifier.
 RUN --mount=type=cache,id=mariadb-apk-${TARGETARCH},sharing=locked,target=/var/cache/apk \
     apk add \
-        mariadb \
-        mysql-client \
```
joecorall, line 12: I removed mysql-client since it's in the base image

postgresql/Dockerfile
```diff
@@ -5,11 +5,14 @@ ARG TARGETARCH
...
 
+ENV \
+    # renovate: datasource=repology depName=alpine_3_20/postgresql16
+    POSTGRESQL_VERSION=16.6-r0 
+
 # Platform specific does require arch specific identifier.
 RUN --mount=type=cache,id=postgresql-apk-${TARGETARCH},sharing=locked,target=/var/cache/apk \
     apk add \
-        postgresql \
-        postgresql-client \
```
joecorall, line 12: I removed `postgresql-client` since it's in the base image

nginx/Dockerfile
```diff
@@ -21,39 +21,46 @@ RUN --mount=type=cache,id=download-downloads-${TARGETARCH},sharing=locked,target
...
-        php83 \
-        php83-ctype \
-        php83-curl \
-        php83-dom \
-        php83-fileinfo \
-        php83-fpm \
-        php83-gd \
-        php83-iconv \
-        php83-intl \
-        php83-json \
```
joecorall, line 38: `php83-json` doesn't appear to be a package

nigelgbanks: Used to be its own package now it resolves to `php83-common` https://pkgs.alpinelinux.org/package/v3.20/community/x86_64/php83-common

base/Dockerfile
```diff
@@ -21,28 +21,64 @@ LABEL License="MIT License"
 # Start s6
 ENTRYPOINT [ "/init" ]
 
+ENV \
```
nigelgbanks, line 24: These would be better exposed as build arguments rather than environment variables, as setting them at runtime has no effect. Additionally, using `ENV` adds a layer to the image.

nginx/Dockerfile
nigelgbanks: Safe to ignore cause `php83-iconv`, etc depend on `php83-common`

base/Dockerfile
```diff
@@ -21,28 +21,64 @@ LABEL License="MIT License"
...
+  # renovate: datasource=repology depName=alpine_3_20/bash
+  BASH_VERSION=5.2.26-r0 \
+  # renovate: datasource=repology depName=alpine_3_20/curl
```
nigelgbanks, line 27: We'll have to remember to update `alpine_3_20` when we change the base `alpine` layer, though I suspect builds will fail if we reference older versions of packages. So I doubt we'll run into a discrepancy, since builds will fail.

nigelgbanks (changes requested):
If `ENV` becomes `ARG` I'm happy to merge this as is. Good stuff!

base/Dockerfile
joecorall: Good call. Fixed with https://github.com/Islandora-Devops/isle-buildkit/pull/380/commits/1b1799abaee6f583e4715dec286decd8ef141845

nigelgbanks (approved):

PR #385, opened 2025-02-14
[skip-release] Automerge apk updates and provide better renovate testing

joecorall:
This PR does four things:

1. When an apk package in isle-buildkit gets an update, auto-merge that if the `test-test/test` CI job succeeds (since we have that set in our branch protection rule as a requirement)
2. Run renovate more often. Previously it was only ran once per week, which was OK when we just checked for dependencies like activemq, composer, etc. But now that package updates are pinned we should get those in sooner than later since a bad pin can break builds
3. Allow running renovate against a PR in this repo. I didn't like that the only way I could easily test changes to our renovate.json was by making changes in the main branch in the repo, so added some config to allow running renovate against a PR
4. Add a new `[skip-release]` option in our PR titles, so the fleet of isle-buildkit containers won't get a patch bump when PRs like this merge that don't effect the actual software running in the containers

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
           TAG=$(autotag)
           git tag $TAG
           git push origin $TAG
-          gh release create $TAG
+          gh release create $TAG --title "$TAG" --generate-notes
           gh workflow run push.yml -f tag=$TAG --ref $TAG
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


joecorall:
Though I **think** this PR might actually create a release since the GHA on the main branch doesn't have these changes 🤔 but the `[skip-release]` should work after this one 😅

nigelgbanks:
Ah looks like this isn't working maybe some difference on the GitHub runners?

error: unknown option '--base-branch'

https://github.com/Islandora-Devops/isle-buildkit/actions/runs/13344960773

Issue #389, opened 2025-02-24
Cantaloupe cannot load turobojpeg library


jonasengelmann:
Thanks for offering your cantaloupe docker image. 

I am experiencing an issue though when loading jpg images:

`Failed to initialize TurboJpegProcessor (error: Can't load library: /opt/libjpeg-turbo/lib/libturbojpeg.so)`

It seems like the symbolic link is broken:
```
# file /opt/libjpeg-turbo/lib/libturbojpeg.so
/opt/libjpeg-turbo/lib/libturbojpeg.so: broken symbolic link to /usr/lib/libturbojpeg.so.0
```

jonasengelmann:
Thanks for the quick fix! I just tested it but now I get this error: 

`Failed to initialize TurboJpegProcessor (error: 'org.libjpegturbo.turbojpeg.TJScalingFactor[] org.libjpegturbo.turbojpeg.TJ.getScalingFactors()')`

joecorall:
Thanks - I can reproduce locally. I am seeing some similar issues e.g. https://github.com/cantaloupe-project/cantaloupe/issues/524

I think there may be a problem with us using libjpegturbo 3.x when the cantaloupe docs specify some specific 2.x version. I'll try building with that 2.x version and see if that resolves the issue.

joecorall:
@jonasengelmann - can you please pull `islandora/cantaloupe:issue-389-libjpeg` and test again? I believe it's fixed by pinning to libjpeg 2.x

jonasengelmann:
Just tested it and it works great:
` e.i.l.c.p.ProcessorConnector - File -> StreamProcessor connection between FilesystemSource and TurboJpegProcessor`

Many thanks for the quick fix!

joecorall:
@jonasengelmann - the PR has been merged and the fix is is now on the `islandora/cantaloupe:main` and any tag above `3.5.7`. Thanks again for reporting the issue!

PR #390, opened 2025-02-24
Pin libjpegturbo to 2.x

joecorall:
Closes https://github.com/Islandora-Devops/isle-buildkit/issues/389

The cantaloupe image was installing libjpeg-turbo with apk, which installs libjpeg-turbo 3.x. However, that library wasn't getting loaded in cantaloupe, so cantaloupe was falling back to `Java2dProcessor` for jpgs

```
Failed to initialize TurboJpegProcessor (error: Can't load library: /opt/libjpeg-turbo/lib/libturbojpeg.so)
```

Once the package was installed properly, we discovered cantaloupe does not work with 3.x. With libjpeg-turbo 3.x installed correctly in cantaloupe it was throwing another error

```
Failed to initialize TurboJpegProcessor (error: 'org.libjpegturbo.turbojpeg.TJScalingFactor[] org.libjpegturbo.turbojpeg.TJ.getScalingFactors()')
```

So I rolled back to `2.0.2` [as the cantaloupe docs recommend](https://cantaloupe-project.github.io/manual/5.0/processors.html#TurboJpegProcessor) and saw the errors went away. Then bumped to the latest version on the 2.x branch: `2.1.5.1` and see cantaloupe loaded it OK

```
20:14:52.075 [qtp951050903-22] DEBUG e.i.l.c.p.ProcessorFactory - TurboJpegProcessor selected for format JPEG (ManualSelectionStrategy offered TurboJpegProcessor, Java2dProcessor)
```

### DIFF
--- a/cantaloupe/Dockerfile
+++ b/cantaloupe/Dockerfile
@@ -46,7 +46,9 @@ RUN --mount=type=cache,id=cantaloupe-apk-${TARGETARCH},sharing=locked,target=/va
         ffmpeg=="${FFMPEG_VERSION}" \
         openjpeg-tools=="${OPENJPG_TOOLS_VERSION}" \
         libjpeg-turbo=="${LIBJPEG_TURBO_VERSION}" \
+        libturbojpeg=="${LIBJPEG_TURBO_VERSION}" \
     && \
+    test -e /usr/lib/libturbojpeg.so.0 && \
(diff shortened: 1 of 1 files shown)


joecorall:
hmm linking to the wrong file. Not seeing it in the build

PR #411, opened 2025-04-06
[skip-release] have renovate check downloads.apache for updates

joecorall:


joecorall:
since this only effects renovate dependency checking, and we can only really test once the config is on the main branch, I'm going to go ahead and merge this.

PR #415, opened 2025-04-17
[skip-release] exclude repology from weekly updates

joecorall:
exclude repology from weekly updates so their updates can be applied immediately. This is to allow our repology rules, which create a PR immediately when the dependency is discovered, to continue working


joecorall:
Going to merge this b/c it can't really be tested until it's in the main branch, and it does not effect our released software

PR #417, opened 2025-04-23
[nginx] include mjs mimetype

joecorall:
Since nginx does not include it in its default mime.types file


joecorall:
This PR is failing because the vscode marketplace moved some time since the code server version we're on, and docker caching has hidden this issue until whenever that happened until now. https://github.com/Islandora-Devops/isle-buildkit/pull/419 should fix it

nigelgbanks (approved):

PR #419, opened 2025-04-24
bump code-server to 4.99.3 and have renovate track updates

joecorall:


joshdentremont:
Looks like maybe we need to update nodejs (https://github.com/Islandora-Devops/isle-nodejs) and then update this: https://github.com/Islandora-Devops/isle-buildkit/blob/af5e5e35b0fdb54f7a282e8645f8301c4df962b5/docker-bake.hcl#L314

based on this: https://github.com/Islandora-Devops/isle-buildkit/actions/runs/14641111484/job/41083721045?pr=419#step:6:312

nigelgbanks:
I pushed up a fix to use the right version of NodeJS.

PR #420, opened 2025-04-30
[minor] add grok processor to cantaloupe

joecorall:


joecorall:
qemu must be slowing down this build. On my local arm machine it only takes like 5-10 min to build this. Gotta rethink this

nigelgbanks:
At some point I would like to use cross compilation for all the tooling, just haven't had the time as of late.

joecorall:
> At some point I would like to use cross compilation for all the tooling, just haven't had the time as of late.

Hmm so if we replaced qemu with instead running the docker builds on separate arm and amd64 github actions runners, have the platform-specific builds push to docker hub with a platform tag, then ran a new GHA job that stitches the cross-platform manifest as our main tag from the two platform-specific images we could have this PR use this new approach, and perhaps we could build everything in this repo instead of the separate nodejs/imagemagick/leptonica repos we have?

nigelgbanks:
Definitely that would be the case. 

I mentioned a re-write I was working on, like 2 months back. I still have the work locally, just not the time to put into it, though hopefully that will change in the next few days. 

This was something I wanted to do on Digital Ocean, since the foundation already has an account there, and we could do it for cheap.  Turns out they don't have ARM backed servers, so we'd pretty much have to do this on AWS. So we'll need to do some convincing for the foundation to set up an account and to approve the costs. Though I'm sure we can do if for like 10-20 a month, likely less.

nigelgbanks:
Oh I didn't realize github had added arm runners... https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

nigelgbanks:
Ah we might be able to do this for free :D

nigelgbanks (approved):

PR #423, opened 2025-05-06
Platform specific runners

joecorall:
As seen in https://github.com/Islandora-Devops/isle-buildkit/pull/420 - using qemu to build multi-platform images takes considerably longer than building for the native CPU architecture. This PR is an attempt to use GitHub's amd+arm runners to build platform-specific images, and a second job to merge those images into a single manifest

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,5 @@
 name: Build
+
 on:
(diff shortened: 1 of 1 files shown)


nigelgbanks (approved):
Looks good to me! Great work!

PR #425, opened 2025-05-06
bump alpine to 3.20.6 and build apk packages in this repo

joecorall:
With https://github.com/Islandora-Devops/isle-buildkit/pull/423 we now can build images faster in GitHub Actions.

So basically cloned two repos we have that build apk packages into their own directories in this repo

- https://github.com/Islandora-Devops/isle-imagemagick -> imagemagick
- https://github.com/Islandora-Devops/isle-leptonica -> leptonica

Then updated docker-bake.hcl and our GHA to build those images.


joecorall:
@nigelgbanks - if we were to do this, do you think nodejs would only trigger a rebuild (which takes ~2h) when something in the image gets bumped?

nigelgbanks:
@joecorall yup it would, perhaps it's time to drop code-server? We should make a post on slack to see if it's actively used?

nigelgbanks:
The caching is a bit hit or miss currently, I'm actively trying some new features to get it to be better. That being said, GitHub only has a local cache of 10GB for the whole repo (not per branch), and the image cache on docker hub doesn't capture the c++ compilation, it's only cached locally. So any time it needs to rebuild it will take forever.

nigelgbanks:
https://islandora.slack.com/archives/CM6F4C4VA/p1747072958771069 I started a thread

joecorall:
While we decide the future of code-server I went ahead and dropped nodejs from this PR to avoid the long build times. We maybe could just keep nodejs as-is in a separate repo. I suspect if we dropped codeserver that will take a lot of effort around documentation, isle-site-template, isle-dc, etc.

nigelgbanks:
So far, the slack makes it look like no one uses it 🤷

joecorall:
If we do end up removing it, which looks like we will, that will probably involve:

- communicate to the community about this changes
- remove code server references from our repos
  - [Islandora-Devops/isle-dc](https://github.com/Islandora-Devops/isle-dc?tab=readme-ov-file#code-server)
  - [Islandora-Devops/isle-site-template](https://github.com/Islandora-Devops/isle-site-template/blob/main/README.template.md#drupal-development)
  - [islandora/documentation](https://github.com/search?q=repo%3AIslandora%2Fdocumentation+%22+ide+%22&type=code) 
- a major version release in this PR

i don't think that work needs to block this PR

nigelgbanks (approved):

PR #426, opened 2025-05-06
[minor] Remove CI docker bake group

joecorall:
Since we split the multi-platform builds across runners. This is a minor bump because I deleted the `4.1.0` release and tag since cantaloupe was failing to build.


joecorall:
merging this since main is broken right now

PR #431, opened 2025-05-14
solr v10 readiness: add SOLR_MODULES

joecorall:
Bug report from @rbos

> This is weird - I had solr fail to load the ISLANDORA core with Error loading class 'solr.ICUCollationField' which indicates that the modules are failing to load properly. Module loading from solrconfig.xml is deprecated as of 9.8, which we're running, but -Dsolr.config.lib.enabled=true is enabled in the solr container.
On a hunch I used the new method and passed SOLR_MODULES with a list of all the available modules, and the core loaded.
So I'm not sure why it stopped working or why it's ignoring solr.config.lib.enabled all of a sudden

Solr 10 will require this, too. Downloading a solr config from drupal results in these comments in the conf:

```
-  <lib dir="${solr.install.dir:../../../..}/modules/extraction/lib" regex=".*\.jar" />
+  <!-- <lib/> directives are deprecated and will be removed in Solr 10.0.
+Ensure to load the required module in your Solr server, for example by appending it to the comma-separated module list enviroment variable like SOLR_MODULES="${SOLR_MODULES},extraction/" -->
+<!-- <lib dir="${solr.install.dir:../../../..}/modules/extraction/lib" regex=".*\.jar" /> -->

-  <lib dir="${solr.install.dir:../../../..}/modules/langid/lib/" regex=".*\.jar" />
+  <!-- <lib/> directives are deprecated and will be removed in Solr 10.0.
+Ensure to load the required module in your Solr server, for example by appending it to the comma-separated module list enviroment variable like SOLR_MODULES="${SOLR_MODULES},langid/" -->
+<!-- <lib dir="${solr.install.dir:../../../..}/modules/langid/lib/" regex=".*\.jar" /> -->

-  <lib dir="${solr.install.dir:../../../..}/modules/ltr/lib/" regex=".*\.jar" />
+  <!-- <lib/> directives are deprecated and will be removed in Solr 10.0.
+Ensure to load the required module in your Solr server, for example by appending it to the comma-separated module list enviroment variable like SOLR_MODULES="${SOLR_MODULES},ltr/" -->
+<!-- <lib dir="${solr.install.dir:../../../..}/modules/ltr/lib/" regex=".*\.jar" /> -->

-  <lib dir="${solr.install.dir:../../../..}/modules/analysis-extras/lib/" regex=".*\.jar" />
+  <!-- <lib/> directives are deprecated and will be removed in Solr 10.0.
+Ensure to load the required module in your Solr server, for example by appending it to the comma-separated module list enviroment variable like SOLR_MODULES="${SOLR_MODULES},analysis-extras/" -->
+<!-- <lib dir="${solr.install.dir:../../../..}/modules/analysis-extras/lib/" regex=".*\.jar" /> -->
```

### DIFF
--- a/images/base/Dockerfile
+++ b/images/base/Dockerfile
@@ -25,7 +25,7 @@ ARG \
   # renovate: datasource=repology depName=alpine_3_22/bash
   BASH_VERSION=5.2.37-r0 \
   # renovate: datasource=repology depName=alpine_3_22/ca-certificates
-  CA_CERTIFICATES_VERSION=20250911-r0 \
+  CA_CERTIFICATES_VERSION=20260413-r0 \
   # renovate: datasource=repology depName=alpine_3_22/curl
   CURL_VERSION=8.14.1-r2 \
   # renovate: datasource=repology depName=alpine_3_22/git
--- a/images/drupal/rootfs/etc/islandora/utilities.sh
+++ b/images/drupal/rootfs/etc/islandora/utilities.sh
@@ -553,7 +553,7 @@ function generate_solr_config {
 
 # Creates a SOLR core for the site using the Solr REST API.
 function create_solr_core {
-    local site core host port
+    local site core host port status
     site="${1}"
     shift
     core=$(drupal_site_env "${site}" "SOLR_CORE")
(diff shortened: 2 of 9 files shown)


rbos:
+1 vote for this, ran into this issue today. Can probably remove `-Dsolr.config.lib.enabled=true` as well if this is specified.

joecorall:
@rbos I think I found out why this was causing issues and was difficult to reproduce. You'll see in this PR I updated our utilities.sh to specify the exact solr version we're using. The [search_api_solr drupal module](https://www.drupal.org/project/search_api_solr) had different behavior whether passing `9` vs `>= 9.8.0` when generating the solr conf.

In prod or a running site going through the admin UI to generate the conf that is a different behavior than what we run in drush and caused the `<lib>` definitions in `solrconfig.xml` to get commented out (which is a breaking change, not sure why they chose to do that when it's just deprecated until solr v10). So if after doing that `SOLR_MODULES` wasn't set it likely caused the break.

This PR should fix it.

PR #435, opened 2025-06-02
Add Transkribus microservice for HOCR derivatives

chryslovelace:
Adds an image for the [transkribus-process](https://github.com/ulsdevteam/transkribus-process) microservice, which encapsulates the third-party [transkribus](https://www.transkribus.org/) API and uses it to generate HOCR derivatives. By default it is disabled in Alpaca.


nigelgbanks (changes requested):
If this could be changed to work as the other images do off base, and utilizes our build of ImageMagick, I don't see any reason for it not to go in. 

Since this requires, from what I can tell, a paid third-party service. Should there be some method to pass on the authentication credentials on to authenticate with transkribus?

Thanks for the contribution!

chryslovelace:
The credentials can be passed as secrets or environment variables to the container, I'll note that in the readme and also refactor this to work off the base image.

nigelgbanks:
Just to let you know, I've seen that you've updated the pull request. I've taken a look and I've seen a number of things that are needed to bring it inline with the rest of the code base. I've added them except for one small issue I seen with persistent data, that I've not yet resolved (I'm learning the ASPNET quirks). I'll sort that tomorrow and raise a pull including your changes, so you don't have to wait much longer.

PR #442, opened 2025-07-05
use native drush uri variable

t4k:
As [noted in Slack](https://islandora.slack.com/archives/CM6F4C4VA/p1751663503305519), it would be helpful to use an environment variable name that Drush can actually use for the URI.


t4k:
See https://github.com/Islandora-Devops/isle-site-template/pull/103.

joecorall (changes requested):
I am +1 on this change, but I think we need to set `DRUSH_OPTIONS_URI` with the value of `DRUPAL_DRUSH_URI` if `DRUSH_OPTIONS_URI` is not set so we don't break existing installs

joecorall (approved):
upon further review, isle-dc is using `DRUSH_OPTIONS_URI`, so this is good to merge

PR #443, opened 2025-07-08
Permit custom S3 endpoint

xurizaemon:
Testing out FCrepo with a non-AWS S3 store, some changes that appear required.

- [ ] Support optional `FCREPO_S3_ENDPOINT`
- [ ] Document `FCREPO_S3_ENDPOINT`

Draft, because I'm just trying it out at this point!


nigelgbanks (approved):
Looks safe as is.

PR #491, opened 2025-11-14
Change source-mime-ext to source-ext in scyllaridae.yml

ajstanley:
This brings the file extension instead of the mimetype into the command, so no more level=ERROR msg="Error building command" err="unknown mime extension: image/tiff" messages.

You can test this by triggering either OCR or HOCR action


joecorall (changes requested):
did you test this? A quick test on my end shows no difference, though I can replicate the problem. I believe the issue came about when this got moved to alpine and doesn't have the stock mimetypes. I am working on a PR

PR #517, opened 2026-01-07
Add MAX_WIDTH to mergepdf [minor]

joecorall:
@aOelschlager noted the 1000x max width was making for some blurry images. So lets make the default 2000px and make it configurable


aOelschlager (approved):
Looks good to me. Thank you!

PR #520, opened 2026-01-14
Revert "Remove UID hack" and make UID mapping more robust

joecorall:
Reverts Islandora-Devops/isle-buildkit#519

I did a bad bit of debugging in https://github.com/Islandora-Devops/isle-site-template/pull/98. Turns out the `UID` was being set with a newline character in it, causing `nginx` to not get its UID updated (and then permissions to fail inside the container). So lets put the DEVELOPMENT_ENVIRONMENT UID swapping back for easier editing on host machines

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -40,7 +40,7 @@ ARG \
   MARIADB_CLIENT_VERSION=11.4.8-r0 \
   # renovate: datasource=repology depName=alpine_3_22/mysql-client
   MYSQL_CLIENT_VERSION=11.4.8-r0 \
-  # renovate: datasource=repology depName=alpine_3_22/netcat-openbsd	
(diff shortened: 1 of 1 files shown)


base/Dockerfile
```diff
@@ -40,7 +40,7 @@ ARG \
   MARIADB_CLIENT_VERSION=11.4.8-r0 \
   # renovate: datasource=repology depName=alpine_3_22/mysql-client
   MYSQL_CLIENT_VERSION=11.4.8-r0 \
-  # renovate: datasource=repology depName=alpine_3_22/netcat-openbsd
+  # renovate: datasource=repology depName=alpine_3_22/netcat-openbsd
```
joecorall:
```suggestion
  # renovate: datasource=repology depName=alpine_3_22/netcat-openbsd
```

PR #526, opened 2026-01-18
[skip-release] Add terraform for configuring cache storage

nigelgbanks:
The spaces exist in Digital Ocean, and this plan doesn't need to be run as part of our CI. Though it's good to have it configured through code. It will be up to developers to maintain and run it if there needs to be a change, though that is very unlikely. 

With this in place, we'll be able to start caching c++ compilation for much faster builds for ImageMagick and similar images.

### DIFF
--- a/terraform/.gitignore
+++ b/terraform/.gitignore
@@ -1,0 +1,24 @@
+# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+
+# Crash log files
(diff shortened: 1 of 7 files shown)


nigelgbanks:
> Can we include in this PR how they're going to be used?

It'll come with the changes that make use of the buckets. It's difficult to document something that doesn't exist in the codebase yet.

> And where do we store secrets?

In Bitwarden, it's documented in the README.md

PR #572, opened 2026-05-14
Update rewrite rule for Drupal in nginx config

joecorall:


images/drupal/rootfs/etc/nginx/shared/drupal.defaults.conf
```diff
@@ -58,7 +58,7 @@ location / {
 }
 
 location @rewrite {
-    rewrite ^/(.*)$ /index.php?q=$1;
+    rewrite ^/(?<path>.*)$ /index.php?q=$path last;
```
joecorall:
```suggestion
    rewrite ^/(?<path>.*)$ /index.php?q=$path;
```

aOelschlager (approved):
I just tested this and it works.

